PR #28, opened 2024-07-01
Add few additional packages

jaudiger:
Add:
- dust
- oha
- tokei
- xsv
- zoxide

### DIFF
--- a/packages/oha/project.bri
+++ b/packages/oha/project.bri
@@ -2,7 +2,7 @@ import * as std from "std";
 import { cargoBuild } from "rust";
 
 export const project = {
-  name: "dust",
+  name: "oha",
   version: "1.4.5",
 };
 
--- a/packages/oha/project.bri
+++ b/packages/oha/project.bri
@@ -1,0 +1,24 @@
+import * as std from "std";
+import { cargoBuild } from "rust";
+
+export const project = {
+  name: "dust",
+  version: "1.4.5",
+};
+
+const crate = std
+  .download({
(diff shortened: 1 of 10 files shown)


packages/oha/project.bri
```diff
@@ -0,0 +1,24 @@
+import * as std from "std";
+import { cargoBuild } from "rust";
+
+export const project = {
+  name: "dust",
```
kylewlacy, line 5: Typo

```suggestion
  name: "oha",
```

kylewlacy (commented):
I still need to test these first, but the code looks good overall besides one small typo! (there should really be a lint or something for that 😅)

packages/oha/project.bri
jaudiger: Aie 😶‍🌫️ my bad, it was a copy and paste... Definitively, a lint rule could help to mitigate these issues. Does the LSP expose any lints? I didn't take a look yet.

jaudiger: Ok, I see: `brioche check` being called in one of a CI job.

kylewlacy: Yeah, currently there's lints from ESLint and typescript-eslint, but there aren't any Brioche-specific lint rules in the LSP yet. I went with ESLint because it has a pretty good plugin system, so hopefully it isn't too hard to start adding lint rules like this as they come up!

kylewlacy (approved):

kylewlacy:
Tested all the new packages, looks good!

PR #29, opened 2024-07-01
Update rust to 1.79.0

jaudiger:
Nothing else, except a really really small refactor to use package.version variable.


kylewlacy (approved):
:+1:

PR #38, opened 2024-07-04
Add new rust packages

jaudiger:
Add:
- Bat
- Broot
- Gitui
- Joshuto


kylewlacy (changes requested):
Hmm, I tried building all of these locally, and I got this error in the gitui build:

```
error: failed to run custom build command for `gitui v0.26.3 (/home/brioche-runner-51d2b1e2f19334c71fec37d2493c9bade0734c0120833017945c3a201544e4d1/work)`

Caused by:
  process didn't exit successfully: `/home/brioche-runner-51d2b1e2f19334c71fec37d2493c9bade0734c0120833017945c3a201544e4d1/work/target/release/build/gitui-8e24a6115633ed1a/build-script-build` (exit status: 101)
  --- stderr
  thread 'main' panicked at build.rs:24:5:
  Can not get git commit: No such file or directory (os error 2)
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Based on the error message, I'm guessing the gitui build process tries to get the commit hash at build time. That means the culprit is likely the `.tar.gz` source of the build, which doesn't include the git repo itself.

I think that one will need to be solved by adding a new `git` package first, along with a `cloneRepo` function that takes a repo and a commit hash and checks it out. I'll open an issue to track this work (there's a bit of nuance in writing the `cloneRepo` function properly)

jaudiger:
Wouah, I'm a bit of surprise here! Locally, I was able to build it:

```bash
> brioche install -p packages/gitui/
Build finished, completed (no new jobs) in 2.17s
Writing output
Wrote output to /root/.local/share/brioche/installed
Note: installation directory not detected in $PATH! Consider adding it:
  /root/.local/share/brioche/installed/bin
```

```bash
> /root/.local/share/brioche/installed/bin/gitui --version
gitui nightly 2024-07-03 ()
```

I wonder why we're not having the same behavior here. But I think I might know, my devenv does not embed git. So, the command in this build script could have been not executed (https://github.com/extrawurst/gitui/blob/master/build.rs).

kylewlacy:
> I wonder why we're not having the same behavior here. But I think I might know, my devenv does not embed git. So, the command in this build script could have been not executed (https://github.com/extrawurst/gitui/blob/master/build.rs).

huh, that's still baffling to me... I have git installed locally, but the build gets run in a container so whether or not git is installed / available locally shouldn't impact the build. The output status isn't checked, so it looks like it will fail if the `git` command isn't found... meaning it looks like the `git` command somehow _was_ found when you built it? Not really sure how to explain that

kylewlacy (approved):

jaudiger:
I can't explain either, on my side I'm sure the git command is not installed in my docker container:

```bash
[root@bb47e0028ffe workspace]# git
bash: git: command not found
[root@bb47e0028ffe workspace]# find / -name git
/root/.local/share/brioche/process-temp/01J1X0CC38RK8F9XQKYQN7ENF7/root/home/brioche-runner-f4abbe2f9e9591a1b125ac4143e1e8556e9b9255f38b23aa518ecdb47cacda6a/work/vendor/vergen/src/feature/git
/usr/lib/node_modules_20/npm/node_modules/@npmcli/git
/workspace/.git/git
[root@bb47e0028ffe workspace]# ls /usr/lib/node_modules_20/npm/node_modules/@npmcli/git
LICENSE  lib  package.json
[root@bb47e0028ffe workspace]# ls /root/.local/share/brioche/process-temp/01J1X0CC38RK8F9XQKYQN7ENF7/root/home/brioche-runner-f4abbe2f9e9591a1b125ac4143e1e8556e9b9255f38b23aa518ecdb47cacda6a/work/vendor/vergen/src/feature/git
cmd.rs  git2.rs  gix.rs  mod.rs
[root@bb47e0028ffe workspace]# ls /usr/lib/node_modules_20/npm/node_modules/@npmcli/git/lib
clone.js  errors.js  find.js  index.js  is-clean.js  is.js  lines-to-revs.js  make-error.js  opts.js  revs.js  spawn.js  utils.js  which.js
```

PR #46, opened 2024-07-05
feat: Add tcsh package

jaudiger:
As described in https://linuxfromscratch.org/blfs/view/stable/postlfs/tcsh.html


kylewlacy (approved):
:+1:

PR #49, opened 2024-07-05
feat: add lurk package

jaudiger:
Add [lurk](https://github.com/JakWai01/lurk) package, which is basically a rust alternative to strace. I tried this package while working on this issue #45


kylewlacy (approved):
Yesss!! using Brioche to build tools to debug Brioche!! I love this

PR #54, opened 2024-07-10
feat: add buildParams and packagePath options to GoBuildOptions interface

jaudiger:
The commit adds new options to the `GoBuildOptions` interface in the `go` package. The `buildParams` option allows specifying build parameters such as running `go generate` before building and passing ldflags to the `go install` command. The `packagePath` option allows setting the package path to build. This change enhances the flexibility and customization of the `goInstall` function.

### DIFF
--- a/packages/go/project.bri
+++ b/packages/go/project.bri
@@ -45,8 +45,15 @@ export function go(): std.Recipe<std.Directory> {
 }
 export default go;
 
+interface GoBuildParameters {
+  generate?: boolean;
+  ldflags?: string[];
+}
+
 interface GoBuildOptions {
   goModule: std.AsyncRecipe<std.Directory>;
+  buildParams?: GoBuildParameters;
+  packagePath?: string;
   runnable?: string;
 }
 
@@ -58,6 +65,10 @@ interface GoBuildOptions {
  *
  * - `goModule`: The Go module to build. Should include `go.mod`, as well as
  *   `go.sum` if external dependencies are needed.
+ * - `buildParams`: Optional build parameters:
+ *   - `generate`: Run `go generate` before building.
+ *   - `ldflags`: An array of ldflags to pass to the `go install` command.
+ * - `packagePath`: Optionally set the package path to build (e.g. `./cmd/foo`).
  * - `runnable`: Optionally set a path to the binary to run
  *   by default (e.g. `bin/foo`).
  *
@@ -69,6 +80,14 @@ interface GoBuildOptions {
  * export default function {
  *   return goInstall({
  *     goModule: Brioche.glob("**\/*.go", "go.mod", "go.sum"),
+ *     buildParams: {
+ *       generate: true,
+ *       ldflags: [
+ *         "-s",
+ *         "-w"
+ *       ],
+ *     },
+ *     packagePath: "./cmd/hello",
  *     runnable: "bin/hello",
  *   });
  * };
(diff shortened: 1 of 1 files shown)


packages/go/project.bri
```diff
@@ -80,13 +99,20 @@ export async function goInstall(
   const modules = goModDownload(options.goModule);
 
   let buildResult = std.runBash`
-    go install
+    # Run generate if requested
+    if [ "${options.buildParams?.generate}" = "true" ]; then
```
kylewlacy, line 103: Nitpick: I'd feel a little better if this were passed as an env var, meaning the code would be `if [ "$go_generate" = "true" ]`. I view it the same way as interpolating a string in a SQL query: it'll usually work and in some cases it will always work, but (in my opinion) it's better practice be consistent and _always_ pass inputs as variables

(eventually I think I might go as far as having a lint warning when interpolating when using `runBash`)

packages/go/project.bri
```diff
@@ -80,13 +99,20 @@ export async function goInstall(
...
+    fi
+
+    go install -ldflags="$ldflags" $package_path
   `
     .workDir(options.goModule)
     .dependencies(go())
     .env({
       GOMODCACHE: modules,
       GOBIN: std.tpl`${std.outputPath}/bin`,
+      ldflags: options.buildParams?.ldflags?.join(" ") ?? "",
```
kylewlacy, line 114: very small and subtle issue, but this won't work correctly if any arg in `ldflags` has spaces in it. According to the Go docs (https://go.dev/cmd/go/?m=old):

> The -asmflags, -gccgoflags, -gcflags, and -ldflags flags accept a space-separated list of arguments to pass to an underlying tool during the build. To embed spaces in an element in the list, surround it with either single or double quotes. The argument list may be preceded by a package pattern and an equal sign, which restricts the use of that argument list to the building of packages matching that pattern [...]

I think there are two directions that would make sense to me:

1. Change `ldflags` to a single string, and let the user worry about argument escaping and the package name
2. Keep `ldflags` as an array, and ensure it gets passed properly. This would mean wrapping args in single quotes or double quotes as needed (and there are of course corner cases when an argument has single quotes, double quotes, and an equals sign)
   - A short-term option would be to just add an assert that fails if any argument contains `'`/`"`/`=` to leave it as a future to-do

kylewlacy (commented):
A few very minor notes on the implementation

packages/go/project.bri
jaudiger: I thought for this one that since we control the value (which is just a primitive boolean) I could have passed it directly, but I'll update the bash to add another env var.

jaudiger: I updated it in the last commit.

packages/go/project.bri
jaudiger: I updated it in the last commit by adding a wrapper to the array containing the ldflags. Basically, I'm escaping single and double quotes as you suggested, but not the =. I had some issues when I tried to escape this special character:

```bash
[58353] /home/brioche-runner-fa2012085166c82d3f8df96aec04cca374bf4df217df39a37b78d082950263da/.local/share/brioche/locals/761f9df90da51dbae24c4c20e4c8ce338a35604cb39c
e4fa611c6c556008d46f/go/pkg/tool/linux_amd64/link: -X flag requires argument of the form importpath.name=value
```

Find below the fixtures I used for the ldflags to verify the escaping was working as expected while building #55:

- ["-s", "-w", \`-X main.version=${project.version}`]
- ["-s", "-w", \`-X 'main.version=${project.version}'`]
- ["-s", "-w", \`-X "main.version=${project.version}"`]
- ["-s", "-w", "-X", \`main.version=${project.version}`]
- ["-s", "-w", "-X", \`'main.version=${project.version}'`]
- ["-s", "-w", "-X", \`"main.version=${project.version}"`]

kylewlacy (approved):
Awesome!

Issue #62, opened 2024-07-12
Support inner directory go.mod files for go mod download


asheliahut:
This is an issue observed when building terraform. `go.mod` files are located in the inner directories for the code base. This can be fixed once [Brioche Issue 79](https://github.com/brioche-dev/brioche/issues/79) is fixed.

kylewlacy:
Resolved in #111

PR #63, opened 2024-07-12
Add in terraform package

asheliahut:
This package relies on updating the go package first to support globbing of `go.mod` files across directories. Once [Brioche Issue 79](https://github.com/brioche-dev/brioche/issues/79) is completed we can then update the go package for issue number #62


kylewlacy:
I fixed the build now that #62 has been resolved! This should be good to merge now

PR #65, opened 2024-07-13
[Package] Normalize install options

jaudiger:
Two commits:

- first one to normalize the usage of `path` in GoBuildOptions interface to make it consistent with other interfaces
- second one to add the ability to set environment variables that could be injected in the `cargo install` command. Some crates rely on these env vars to be customized


kylewlacy (approved):
LGTM, I like the standardized name better here!

Issue #43, opened 2024-07-04
Add Git package


kylewlacy:
Pretty self-explanatory. The `git` executable is obviously just a useful piece of software to install or call.

For Brioche, we'll want at least one non-trivial utility method: cloning a Git repo. GitHub offers source `.tar.gz` downloads for every tag in a repo, but this isn't sufficient for every project. In #38, there was a snag because [gitui](https://github.com/extrawurst/gitui) expected to get information about the source repo at build time, which practically can only be provided by cloning the repo via Git directly.

A while ago, I came across this article, which provides some detailed insight for cloning a Git repo from a single hash: <https://blog.hartwork.org/posts/clone-arbitrary-single-git-commit/>. I think the Brioche `git` package should have a `gitCheckout` function, which takes a repo URL and a commit hash (validated by a regex), then clones just that commit. This is pretty important to get right, as we want the function to be consistent even if the repo adds more commits in the future (meaning that commits ahead of the provided commit shouldn't be in the cloned repo at all)

PR #66, opened 2024-07-14
Add in base for Git Package

asheliahut:
- Fixes Curl first to add in libcurl to the curl package.
- Next adds in Git so future additions can be added to it.


kylewlacy (approved):
LGTM

Part of #43

PR #67, opened 2024-07-14
[Package] Normalize build package options

jaudiger:
Unify the build options to help to better use the current helpers across different languages.


kylewlacy (approved):
Awesome

PR #70, opened 2024-07-16
[Bat] add a unit test

jaudiger:
Add a sort of unit test for `bat` package based on what was done with `openssl`. It helps to figure out if the package is at least built correctly with a basic test.

To run it (not for @kylewlacy since he knows it):

```bash
brioche build -p packages/bat -e test
```

If it suits you (@kylewlacy), I could do the same for the other packages.

### DIFF
--- a/packages/bat/project.bri
+++ b/packages/bat/project.bri
@@ -16,9 +16,23 @@ const source = std
   .unarchive("tar", "gzip")
   .peel();
 
-export default () => {
+export default function bat(): std.Recipe<std.Directory> {
   return cargoBuild({
     source,
     runnable: "bin/bat",
   });
-};
+}
+
+export async function test() {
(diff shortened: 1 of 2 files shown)


kylewlacy:
> If it suits you (@kylewlacy), I could do the same for the other packages.

I'm fine with adding these kinds of tests for one-offs like this (e.g. where the test was added because it was useful), so I'm good to merge this PR, but I'd suggest holding off going through every package for now. At a minimum I'd like to at least have some way to automate the tests in CI/CD first

But ideally, I'd like to try adding a proper `brioche test` subcommand first. I think I want it to be a bit more than just calling `export function test()`. I opened brioche-dev/brioche#94 as a larger issue for `brioche test`, and I'll also start writing down some design ideas there

PR #78, opened 2024-07-23
feat: Add alsa-lib package

jaudiger:
Add [alsa-lib](https://github.com/astral-sh/ruff) tool.

There are two possibilities to get the source of alsa-lib:

- through git: https://github.com/alsa-project/alsa-lib
- through their website: https://github.com/alsa-project/alsa-lib

Following what is stated [here](https://github.com/alsa-project/alsa-lib/blob/master/INSTALL), the tarball archive from the website is already pre-configured to compile it, while the git sources needed another step to set up the build process by calling the script `gitcompile`. I decided to follow what is already done on [Nixpkgs store](https://github.com/NixOS/nixpkgs/blob/nixos-24.05/pkgs/by-name/al/alsa-lib/package.nix#L39) and on [Homebrew store](https://github.com/Homebrew/homebrew-core/blob/5d936da59a39bab63aad4edd3f6835673388bb27/Formula/a/alsa-lib.rb) by using the tarball archive from the website.

I was trying to package [Sniffnet](https://github.com/GyulyVGC/sniffnet), but alsa-lib is a required package.

### DIFF
--- a/packages/alsa_lib/project.bri
+++ b/packages/alsa_lib/project.bri
@@ -1,7 +1,7 @@
 import * as std from "std";
 
 export const project = {
-  name: "alsa-lib",
+  name: "alsa_lib",
   version: "1.2.12",
 };
 
--- a/packages/alsa_lib/project.bri
+++ b/packages/alsa_lib/project.bri
@@ -1,0 +1,32 @@
+import * as std from "std";
+
+export const project = {
+  name: "alsa-lib",
+  version: "1.2.12",
+};
+
+const source = std
+  .download({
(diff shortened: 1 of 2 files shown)


packages/alsa_lib/project.bri
```diff
@@ -0,0 +1,32 @@
+import * as std from "std";
+
+export const project = {
+  name: "alsa-lib",
```
kylewlacy, line 4: Another case where https://github.com/brioche-dev/brioche/issues/67 would help!

```suggestion
  name: "alsa_lib",
```

jaudiger: Indeed!

kylewlacy (approved):
LGTM

PR #81, opened 2024-07-25
feat: Add pcre2 package

jaudiger:
Resolve #45 

Package [pcre2](https://github.com/PCRE2Project/pcre2) which is a common library used by other tools such as `git`, `ripgrep`, etc.

The packaging is a bit hacky due to hard-coded paths in a few build scripts, see the discussion in the related issue #45.

```bash
bash-5.2$ brioche build -p packages/pcre2 -o /tmp/pcre2
Build finished, completed (no new jobs) in 2.76s
Result: 26526a47af9efabb19badc404039cf41aa89874b3bdd1d3bf37f39c743d4715d
Writing output
Wrote output to /tmp/pcre2

bash-5.2$ ls /tmp/pcre2
bin  brioche-env.d  brioche-resources.d  include  lib  share

bash-5.2$ ls /tmp/pcre2/lib/pkgconfig/
libpcre2-16.pc  libpcre2-32.pc  libpcre2-8.pc  libpcre2-posix.pc
```

PR #82, opened 2024-07-25
Add ripgrep package

jaudiger:
Package [ripgrep](https://github.com/BurntSushi/ripgrep).

Finally, I didn't need #81, since pcre2 library was built from the crate https://github.com/BurntSushi/rust-pcre2.

This PR also contains a new way to set cargo build option based on what's already done with go build parameters.


kylewlacy (approved):
:+1:

kylewlacy:
And love the new way to set features in Rust!

PR #146, opened 2024-12-21
[std] Add `std.attachResources()` recipe

kylewlacy:
Companion to brioche-dev/brioche#149

This PR adds a new `std.attachResources()` function. This returns a recipe corresponding to the new upstream `attach_resources` recipe type.


kylewlacy:
This PR should only be merged once the `attach_resources` recipe type has landed in a published Brioche release. The implementation in this PR currently asserts that `std.attachResources()` requires Brioche >= v0.1.4, which should be the first version to include this new recipe.

kylewlacy:
Brioche v0.1.4 went out yesterday, so this should be ready to go now

PR #164, opened 2025-01-10
try filtering packages for checks and update go to 1.23.4

asheliahut:
Also to be able to test the new workflow I needed to make changes to a project, I chose Go
- Go has been updated to 1.23.4


asheliahut:
Closing due to the checks should be done at brioche for the whole workspace. As checking can be needed for all projects if brioche main has changed.

PR #171, opened 2025-01-19
refactor(asciinema): update source retrieval method to use tar.gz archive

jaudiger:
But maybe there was a reason to not having used the method `Brioche.download`.


kylewlacy (approved):
For a while, I ended up defaulting to using git directly over GitHub's source tarballs, but I definitely wasn't consistent with it so I think it makes sense to stick to using `Brioche.download` as the default since that matches what the majority of packages are doing

It might be worth revisiting what the default should be. My fear is that we could hit breakages one day because GitHub [doesn't guarantee that their source tarballs are always byte-for-byte identical](https://github.blog/open-source/git/update-on-the-future-stability-of-source-code-archives-and-hashes/), whereas git hashes should have stronger guarantees. But for now, I think it's the right call to stick with the default and split that out as a separate issue to consider

kylewlacy:
As a follow up, I opened a Discussion around using `Brioche.download()` vs. `Brioche.gitRef()` for downloading sources for GitHub releases here: https://github.com/brioche-dev/brioche-packages/discussions/176

PR #211, opened 2025-01-25
feat: enhance alsa_lib by adding testing and auto-update functionalities

jaudiger:
This is part of the POC for [testing](https://github.com/brioche-dev/brioche/issues/94) and [auto-update](https://github.com/brioche-dev/brioche/issues/165) functionalities.

## Testing

```bash
> brioche build -p packages/alsa_lib/ -e test      
141    │ 1.2.13
 0.87s ✓ Process 141
 8.08s ✓ Registry  100% Fetch 136 blobs + 163 recipes from registry                                                                                                                                                                        
14.73s ✓ Registry  100% Fetch 1 blob + 1 recipe from registry                                                                                                                                                                              
 2.02s ✓ Registry  100% Fetch 1 blob + 1 recipe from registry                                                                                                                                                                              
Build finished, completed 6 jobs in 1m32s
Result: 52e866444decd8be3438780f92a27cbe76dd34b9fea832de3ca29c37b35703eb
```

## Auto-Update

```bash
> brioche build -p packages/alsa_lib/ -e autoUpdate
40     │ 1.2.13
 0.51s ✓ Process 40
 1.65s ✓ Registry  100% Fetch 1 blob + 1 recipe from registry                                                                                                                                                    
 0.63s ✓ Registry  100% Fetch 1 blob + 1 recipe from registry                                                                                                                                                    
 0.62s ✓ Registry  100% Fetch 1 blob + 1 recipe from registry                                                                                                                                                    
Build finished, completed 11 jobs in 20.05s
```

I took the decision to use nushell for the auto-update feature. It enables easier IMO writing of the script to retrieve the latest release for a package. I personally like the workflow of the nushell scripts. But it's a personal opinion. I'm open on this point. But I think we should stick with the same approach for other packages, if we validate this one.

### DIFF
--- a/packages/alsa_lib/project.bri
+++ b/packages/alsa_lib/project.bri
@@ -1,3 +1,5 @@
+import caCertificates from "ca_certificates";
+import nushell from "nushell";
 import * as std from "std";
 
 export const project = {
@@ -11,7 +13,7 @@ const source = Brioche.download(
   .unarchive("tar", "bzip2")
   .peel();
 
-export default function (): std.Recipe<std.Directory> {
+export default function alsaLib(): std.Recipe<std.Directory> {
   const alsaLib = std.runBash`
     ./configure --prefix=/
     make install DESTDIR="$BRIOCHE_OUTPUT"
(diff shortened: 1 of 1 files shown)


packages/alsa_lib/project.bri
```diff
@@ -26,3 +28,57 @@ export default function (): std.Recipe<std.Directory> {
...
+    .process({
+      command: std.tpl`${nushell()}/bin/nu`,
+      args: [src],
+      unsafe: {
+        networking: true,
+      },
+    })
+    .dependencies(caCertificates())
+    .toFile();
+}
```
kylewlacy, line 84: @jaudiger thoughts on this version of `autoUpdate`? There are two tweaks I made:

1. Return the full `project` JSON value instead of just the version. A little forward thinking but eventually I want to be able to support multiple versions in a single package ([brioche-dev/brioche#97](https://github.com/brioche-dev/brioche/issues/97)), where I think the JSON returned would get more complicated. Think like Rust, where you'd want to have like `stable` / `beta` / `nightly` mapped to specific version numbers.
2. Return a runnable recipe that prints the result, instead of returning a recipe containing the result (i.e. you'd do `brioche run -e autoUpdate` instead of `brioche build -e autoUpdate`). This one's subtle, but `brioche build` would get cached forever. If you ran it today with no update, it'd tell you there was no new version then cache the result. Then tomorrow, if there was a new release, it would still have the cache that there was no new version! It'd take a lot of work to change caching to support this, so for MVP at least, basing it on `brioche run` is simpler

```suggestion
export function autoUpdate() {
  const src = std.file(std.indoc`
    let version = http get https://www.alsa-project.org/files/pub/lib
      | lines
      | where {|it| ($it | str contains "alsa-lib") and (not ($it | str contains ".sig")) }
      | parse --regex '<a href="alsa-lib-(?<version>.+)\.tar\.bz2">'
      | sort-by --natural --reverse version
      | get 0.version
    $env.project | from json | update version $version | to json
  `);

  return std.withRunnable(std.directory(), {
    command: "nu",
    args: [src],
    env: { project: JSON.stringify(project) },
    dependencies: [nushell()],
  });
}
```

jaudiger: I made the modifications based on your comments. It was valuable. And I see where you want to go, it makes sense.

kylewlacy (approved):
LGTM 💯 

Seeing the `test` function build a small C program is really cool too!

PR #214, opened 2025-01-28
Update GitHub packages reference

jaudiger:
Prefer usage of gitCheckout() + Brioche.gitRef() instead of downloading directly the tarball file for packages available on Github. See this discussion for more information: https://github.com/brioche-dev/brioche-packages/discussions/176

Two packages have not been migrated:

- git
- openssl

Since both are used for the command gitCheckout(), which make me wonder how we could resolve this circular dependency other than doing the solution (1) of the discussion.

### DIFF
--- a/packages/jq/project.bri
+++ b/packages/jq/project.bri
@@ -1,4 +1,3 @@
-import { gitCheckout } from "git";
 import * as std from "std";
 
 export const project = {
@@ -6,12 +5,11 @@ export const project = {
   version: "1.7.1",
 };
 
-const source = gitCheckout(
-  Brioche.gitRef({
-    repository: "https://github.com/jqlang/jq.git",
-    ref: `jq-${project.version}`,
-  }),
-);
+const source = Brioche.download(
+  `https://github.com/jqlang/jq/releases/download/jq-${project.version}/jq-${project.version}.tar.gz`,
+)
+  .unarchive("tar", "gzip")
+  .peel();
 
 export default function (): std.Recipe<std.Directory> {
   const jq = std.runBash`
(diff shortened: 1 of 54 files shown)


packages/jq/project.bri
kylewlacy, line 1: `jq` is another example like `openssl` where the URL here is a release download rather than a git archive, and one published by the jq project. Trying out `brioche build -p packages/jq` here fails because the upstream source doesn't have `./configure` while the original tarfile does, so I think it'd be better to revert this one

(Longer term, it might be better to always use the autoconf build instead of the pre-packaged `./configure`, but that's a different conversation IMO and shouldn't be in scope for this PR)

kylewlacy (commented):
LGTM overall! Caught one failure in `jq` though

For `openssl`, luckily it seems like the current URL structure we're using is stable. Like jq, it's a download from `/releases/download/*`, not `/tags/archive/*`, meaning it's a tarfile published by OpenSSL directly, not a git archive published by GitHub. Most likely, it too includes a pre-configured `./configure` script generated by autoconf, meaning it wouldn't be a direct mirror for the git repo. This one should continue to be safe to download as-is.

As for `git`... that one is a git archive and I don't have a good answer for it. I don't really have a better plan than "cross our fingers GitHub doesn't change the hash on us", at least for now.

kylewlacy (approved):
Alright, I reverted the commit that updated `jq`, everything left looks good!

packages/jq/project.bri
jaudiger: I find it quite strange that most of the release downloads contain the script `./configure` whereas the git archive does not. I thought we always needed to generate the `./configure` through tools as `autoreconf`. Let's leverage for now the possibility to use the pre-built one that comes with the release download. Indeed it makes things much easier for now.

PR #218, opened 2025-01-29
Fix git repo URL to fix `carapace` build

kylewlacy:
Followup to #214, it seems like the git repo URL `https://github.com/carapace-sh/carapace.git` was used, where it should be `https://github.com/carapace-sh/carapace-bin.git`. I missed that during the review!

Verified locally that the build works with the adjusted repo URL

### DIFF
--- a/packages/carapace/brioche.lock
+++ b/packages/carapace/brioche.lock
@@ -1,8 +1,8 @@
 {
   "dependencies": {},
   "git_refs": {
-    "https://github.com/carapace-sh/carapace.git": {
-      "v1.1.1": "74167cbe2c66c0bbd8a11c5adcd58ec68b1be337"
+    "https://github.com/carapace-sh/carapace-bin.git": {
+      "v1.1.1": "08c3a82ad974e6904329b6b674670689da068d81"
     }
   }
 }
--- a/packages/carapace/project.bri
+++ b/packages/carapace/project.bri
@@ -9,7 +9,7 @@ export const project = {
 
 const source = gitCheckout(
   Brioche.gitRef({
-    repository: "https://github.com/carapace-sh/carapace.git",
+    repository: "https://github.com/carapace-sh/carapace-bin.git",
     ref: `v${project.version}`,
   }),
 );


jaudiger:
I missed that too, sorry !

jaudiger:
Did you detect it with the CI ?

kylewlacy:
...but in the future, if we have more GitHub Actions runners, or faster GitHub Actions runners, or a sponsor helping with costs, etc. then it might make sense to fully automate all PRs. Definitely worth trying to figure out options here since automated checks are super valuable

jaudiger:
Ohh okay. I wasn't aware of the trick with prepending the branch with `build/`. I'll do it next time, I'm updating too many packages in parallel in the same PR.

I see your point. When you say the server is self-hosted ? Does it mean you run it at home, or is hosted somewhere on a cloud provider infrastructure on a small instance ? And what is the current spec of the instance ? Maybe we could discuss that somewhere rather than in this PR 😆 

Well, I was asking for the build inside PR because since I'm using an ARM machine (Apple MacBook Pro M1), I need to virtualize brioche inside a docker image running with Rosetta (that translates instructions between x86 and ARM architectures). It's not that slow, but I can see the difference when building something inside this virtual environment. Thus, testing all the packages by building them can sometimes take time. But I'm trying to do as much as possible. Well, for this PR, I was doing it but on random packages. So, my fault for the faulty carapace URL and for jq building.

I see what you mean, it reminds me of the Nixpkgs repository where they have a specific job to compute the labels of a PR (https://github.com/NixOS/nixpkgs/actions/runs/13040123824/job/36379874774?pr=377814). These labels inform for example on the number of packages impacted by the modification in the PR, and so on. I understand right now that we can't run build on the global directory like that. But, maybe, we could do something like that:
- Run the build on the global directory on main branch
- Run the build on the brioche projects affected by the files modified. It will still be approximate, it won't take into account the dependencies injection. But at least, it's a step in better testing the PR before merging them.

kylewlacy:
(For visibility, I replied to this conversation through Discord: https://discord.com/channels/1246413282894413844/1246601881585975441/1334409958808227880)

PR #225, opened 2025-01-30
Add in gh cli with test/autoupdate

asheliahut:


packages/gh/project.bri
```diff
@@ -0,0 +1,63 @@
...
+  return script;
+}
+
+export function autoUpdate() {
+  const src = std.file(std.indoc`
+    let version = http get https://api.github.com/repos/cli/cli/releases/latest
+      | get tag_name
+      | str substring 1..
+
+    $env.project | from json | update version $version | to json
```
kylewlacy, line 54: #nonblocking Would be kinda cool to use this to also update `latestBuildDate` here too based on the tag version!

kylewlacy (commented):
LGTM

One thing I wanted to suggest before merging: maybe rename to `github_cli`? They seem to be consistent with both the repo name and in the README to call it "GitHub CLI", where `gh` is just the command name (similar to how we went with `jujutsu` instead of `jj`). Thoughts?

packages/gh/project.bri
asheliahut: Sounds good!

packages/github_cli/project.bri
```diff
@@ -0,0 +1,75 @@
...
+  return script;
+}
+
+export function autoUpdate() {
+  const src = std.file(std.indoc`
+    let releaseData = (http get https://api.github.com/repos/cli/cli/releases/latest | from json)
+
+    let version = $releaseData
+      | get tag_name
+      | str replace '^v' ''
```
asheliahut, line 54: To note this seems to be a better method so it can be reused throughout all of the projects vs just cut out the first character. It will be more exact.

jaudiger (approved):

packages/github_cli/project.bri
jaudiger: I like it, indeed, it's better to have a more generic function here

PR #226, opened 2025-02-02
Standardize version string handling by replacing 'v' prefix in auto update

jaudiger:
Per discussion https://github.com/brioche-dev/brioche-packages/pull/225#discussion_r1936235366

### DIFF
--- a/packages/github_cli/project.bri
+++ b/packages/github_cli/project.bri
@@ -51,7 +51,7 @@ export function autoUpdate() {
 
     let version = $releaseData
       | get tag_name
-      | str replace '^v' ''
+      | str replace --regex '^v' ''
 
     let latestBuildDate = $releaseData
       | get created_at
@@ -63,7 +63,6 @@ export function autoUpdate() {
       | update version $version
       | update latestBuildDate $latestBuildDate
       | to json
-
   `);
 
   return std.withRunnable(std.directory(), {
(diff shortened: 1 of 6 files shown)


packages/github_cli/project.bri
```diff
@@ -51,7 +51,7 @@ export function autoUpdate() {
 
     let version = $releaseData
       | get tag_name
-      | str replace '^v' ''
+      | str replace --regex '^v' ''
```
jaudiger, line 54: @asheliahut I had to adjust the argument here to properly use the regex capabilities of `str replace`. Otherwise it was just doing a raw find and replace.

asheliahut: I wonder if there is a way to take this stuff for GitHub and make it generic so we put in the repo name, and get back this part of this and a path for any replacement for things in version.

jaudiger: @kylewlacy began to expose possible solutions here: https://github.com/brioche-dev/brioche-packages/discussions/176 to hide this work into a dedicated function that could be static or not (depending of the option chosen).

asheliahut: Awesome, Yeah I am not saying this needs to be done here. I am good with accepting. Sorry I have been really focused on work lately.

asheliahut (approved):

kylewlacy:
Makes sense to me to merge as-is!

PR #266, opened 2025-03-19
Adding in the basis for php

asheliahut:


kylewlacy (approved):
LGTM

I pushed up a few minor tweaks: using `-j16` to run the build in parallel, removing `$PKG_CONFIG_PATH` (since PHP doesn't seem to have a pkg-config dir), and bumped the package to v8.4.5

PR #271, opened 2025-03-29
Add `llvm` package

kylewlacy:
> Small note about this PR: I tested this build against LLVM v20.1.0, but a new version had released since I started work on this package. I updated the package definition, but due to the very slow build times, I'm going to merge it as-is and let the CI pipeline take care of the build.

This PR adds a new package for [LLVM](https://llvm.org/), a compiler/toolchain project that includes multiple sub-projects including Clang (a C/C++ compiler), LLDB (a debugger), and LLD (a linker).

This PR adds initial support for LLVM, and the default build includes some, but not all, of LLVMs projects: currently, it includes `clang`, `clang-tools-extra`, `lld`, and `lldb`.

I also added an initial iteration of a toolchain function, called `llvm.llvmToolchain()`. It's meant to be similar-ish to `std.toolchain()`, except use LLVM-based tools where possible. Currently, it relies on `std.toolchain()`-- namely, for glibc and standard include headers. I'd like to get it to a point where it's independent from `std.toolchain()` in the future.

I also haven't tested `llvm.llvmToolchain()` much, but here's a minimal working example:

```ts
import * as std from "std";
import { llvmToolchain } from "llvm";

export default function () {
  return std.runBash`
    mkdir -p "$BRIOCHE_OUTPUT/bin"
    clang -v main.c -o "$BRIOCHE_OUTPUT/bin/hello-world"
    ln -s bin/hello-world "$BRIOCHE_OUTPUT/brioche-run"
  `
    .workDir(
      std.directory({
        "main.c": std.file(std.indoc`
          #include <stdio.h>

          int main() {
              printf("Hello, world!\\n");
              return 0;
          }
        `),
      }),
    )
    .dependencies(llvmToolchain())
    .toDirectory();
}
```

### DIFF
--- a/packages/llvm/project.bri
+++ b/packages/llvm/project.bri
@@ -34,6 +34,15 @@ export default function llvm(): std.Recipe<std.Directory> {
     CPATH: { append: [{ path: "include" }] },
   });
 
+  // Remove a bunch of `*.tmp*` files left over in the build
+  // TODO: Figure out where these temp files are coming from!
+  llvm = std.runBash`
+    cd "$BRIOCHE_OUTPUT"
+    find . -name '*.tmp*' -delete
+  `
+    .outputScaffold(llvm)
+    .toDirectory();
+
   return llvm;
 }
 
--- a/packages/llvm/project.bri
+++ b/packages/llvm/project.bri
@@ -47,6 +47,14 @@ export default function llvm(): std.Recipe<std.Directory> {
   return llvm;
 }
 
+/**
+ * A minimal toolchain that uses LLVM-based tools for compilation, such
+ * as Clang and LLD. It currently still uses components from
+ * `std.toolchain()`, such as system libraries and headers.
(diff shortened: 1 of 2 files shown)


packages/llvm/project.bri
```diff
@@ -0,0 +1,110 @@
...
+    LIBRARY_PATH: { append: [{ path: "lib" }] },
+    CPATH: { append: [{ path: "include" }] },
+  });
+
+  // Remove a bunch of `*.tmp*` files left over in the build
+  // TODO: Figure out where these temp files are coming from!
+  llvm = std.runBash`
+    cd "$BRIOCHE_OUTPUT"
+    find . -name '*.tmp*' -delete
+  `
```
kylewlacy, line 43: Still no clue why these leftover temp files are here... maybe it's a result of our custom CMake patches? This should definitely get fixed

PR #326, opened 2025-04-23
refactor(testing): uniformize test functions

jaudiger:
This PR does a few things:

- Always ensure that the `script` variable is cast as a file before being returned
- Remove the `echo -n $()" encapsulation in few test methods, and always trim the empty end of line on typescript side
- Try to always use `result` instead of `output` for script result content.


kylewlacy (approved):
Awesome :100:

PR #333, opened 2025-04-24
feat(nodejs): add test and autoUpdate capabilities

jaudiger:
```bash
[container@ba5998f2ae1e workspace]$ brioche run -e autoUpdate -p packages/nodejs
Build finished, completed (no new jobs) in 2.53s
Running brioche-run
{
  "name": "nodejs",
  "version": "23.11.0"
}
[container@ba5998f2ae1e workspace]$ brioche build -e test -p packages/nodejs
Build finished, completed (no new jobs) in 1.64s
Result: 9e89d218601329b5aac2aa3dab82b9df143651b77642fd580b402f7b8b334b5a
```

PR #334, opened 2025-04-24
chore(autoUpdate): update API endpoint to fetch tags for versioning

jaudiger:
While working on https://github.com/brioche-dev/brioche-packages/pull/333, I was first getting a 504 in the autoUpdate function. During the initial testing of the nodejs PR, doing a GET on `https://api.github.com/repos/nodejs/node/git/matching-refs/` was leading a 504 error code from the GitHub API. After having read the documentation of [this REST API](https://docs.github.com/en/rest/git/refs?apiVersion=2022-11-28#list-matching-references), I saw that it was returning all the heads, tags, remotes of the repositories. While we only want the tags here.

Here is a quick showcase test (using Nushell) to demonstrate the different capabilities of the API:

```nu
➜ http get https://api.github.com/repos/aws/aws-cli/git/matching-refs/tags | length 
2833

➜ http get https://api.github.com/repos/aws/aws-cli/git/matching-refs/heads | length 
85

➜ http get https://api.github.com/repos/aws/aws-cli/git/matching-refs/ | length 
7757
```

The thing is, the larger the output, the longer it takes to process. Only fetching the tags here lead to two advantages:

- Reduce the size of the response (and thus, reduce the possibility to get a 504 from the GitHub API)
- Reduce the process time by stripping during the first stage the unwanted data

### DIFF
--- a/packages/aws_cli/project.bri
+++ b/packages/aws_cli/project.bri
@@ -128,7 +128,7 @@ export async function test() {
 
 export function autoUpdate() {
   const src = std.file(std.indoc`
-    let version = http get https://api.github.com/repos/aws/aws-cli/git/matching-refs/
+    let version = http get https://api.github.com/repos/aws/aws-cli/git/matching-refs/tags
       | get ref
       | each {|ref|
         $ref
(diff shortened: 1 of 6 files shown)


kylewlacy (approved):
Good catch! LGTM

PR #329, opened 2025-04-24
feat(caCertificates): add test and autoUpdate capabilities

jaudiger:
```bash
[container@ba5998f2ae1e workspace]$ brioche run -e autoUpdate -p packages/ca_certificates/
Build finished, completed (no new jobs) in 2.39s
Running brioche-run
{
  "name": "ca_certificates",
  "version": "2025-02-25"
}
[container@ba5998f2ae1e workspace]$ brioche build -e test -p packages/ca_certificates/
Build finished, completed (no new jobs) in 1.62s
Result: 38c26ecd02963d2106b8c2df12707cef2f090789a49f744e7ee0cafd14892328
```

PR #335, opened 2025-04-25
Allow Zip in Brioche typing for archive format 

asheliahut:
Fixes brioche-dev/brioche#235

### DIFF
--- a/packages/std/core/runtime.bri
+++ b/packages/std/core/runtime.bri
@@ -239,7 +239,7 @@ export type Platform = "x86_64-linux" | "aarch64-linux";
 
 export type Hash = { type: "sha256"; value: HexString };
 
-export type ArchiveFormat = "tar";
+export type ArchiveFormat = "tar" | "zip";
 
 export type CompressionFormat = "none" | "bzip2" | "gzip" | "xz" | "zstd";
 


jaudiger (approved):
LGTM, but we should check with @kylewlacy to see if we can proceed with this PR as is, or if we need to wait until earlier versions of Brioche are officially removed (other PRs are currently blocked until Brioche version 0.1.5 is the only supported one, see https://github.com/brioche-dev/brioche-packages/pull/329#issuecomment-2827497412).

My main concern would be to next add a package that rely on the new zip archive, and another person trying to build it with Brioche 0.1.4. I think we can move forward after [2025-04-30](https://brioche.dev/blog/announcing-brioche-v0-1-5/#psa-sunsetting-support-for-older-versions-of-brioche). So, just a few more days.

kylewlacy:
Seconded @jaudiger, we _could_ merge this today but it would cause compatibility issues with Brioche v0.1.4 or earlier.

Other similar features use semver checks to error when used with earlier version ([example in `std.setEnv`](https://github.com/brioche-dev/brioche-packages/blob/a705f6796ce3b013c5e0bf7259ae3bcf836c9221/packages/std/extra/set_env.bri#L57)). We could do the same here, but the problem is that using zip archives would still fail when being uploaded to the registry unless the registry were updated (I think). But since the registry is going to be simplified a lot after 2025-04-30, it'd just be easier to hold off for now then merge afterward

PR #364, opened 2025-04-27
chore: remove unecessary export modifier from source definitions

jaudiger:
This PR standardizes the declaration of `source` variables across projects by removing the `export` keyword. We don't need to export it normally.


kylewlacy (approved):
LGTM

I would like to make it possible to grab sources in the future as a convenience (e.g. `brioche build -r linux -e source -o ./linux-source`), but there's work needed on the Brioche side to enable that, since is expects the export to be a function

jaudiger:
> I would like to make it possible to grab sources in the future as a convenience (e.g. `brioche build -r linux -e source -o ./linux-source`), but there's work needed on the Brioche side to enable that, since is expects the export to be a function

Got it, some persons could be interested about it, indeed.

PR #372, opened 2025-04-28
Add `htop` package

kylewlacy:
[htop](https://htop.dev/) is an interactive TUI process viewer-- basically an interactive version of `top`

### DIFF
--- a/packages/htop/brioche.lock
+++ b/packages/htop/brioche.lock
@@ -1,0 +1,9 @@
+{
+  "dependencies": {},
+  "downloads": {
+    "https://github.com/htop-dev/htop/releases/download/3.4.1/htop-3.4.1.tar.xz": {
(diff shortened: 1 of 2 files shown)


kylewlacy:
This is one of a few TUI tools in Brioche. Trying to run it directly resulted in an error for me:

```sh-session
$ brioche run -p packages/htop
Build finished, completed (no new jobs) in 0.40s
Running brioche-run
Error opening terminal: xterm-256color.
```

After a bit of research, I found out this was due to how (our version of) ncurses looks for terminfo files: it's checking the path `//share/terminfo` by default. To resolve this, I ran it with the `$TERMINFO_DIRS` env var set like this:

```sh
TERMINFO_DIRS=/usr/share/terminfo brioche run -p packages/htop
```

PR #382, opened 2025-04-28
refactor: add missing return type to default functions

jaudiger:
This PR standardizes the return type of default export functions across some project files by explicitly specifying the `std.Recipe<std.Directory>` type.


kylewlacy (approved):
:+1:

PR #423, opened 2025-05-04
[std] Implement `Brioche.gitCheckout()` global function

kylewlacy:
This PR updates `std` to add the global function `Brioche.gitCheckout()`, and additionally replaces most uses of `git.gitCheckout` with `Brioche.gitCheckout`.

`Brioche.gitCheckout` is a new [static](https://brioche.dev/docs/core-concepts/statics/) (documentation pending), and runtime support was added in Brioche v0.1.5. From the runtime's perspective, it behaves just like `Brioche.gitRef()`, so it takes a repository URL and a git ref, and saves the commit of the git ref in the lockfile. But the implementation is just like `git.gitCheckout()`, so it returns a recipe containing a clone of the repository at that commit.

```typescript
// Previously:
// const source = gitCheckout(
//   Brioche.gitRef({
//     repository: "https://github.com/brioche-dev/brioche.git",
//     ref: "main",
//   }),
// );

// Now:
const source = Brioche.gitCheckout({
  repository: "https://github.com/brioche-dev/brioche.git",
  ref: "main",
});
```

Implementation-wise, `Brioche.gitCheckout()` just imports and calls `git.gitCheckout()`! I originally didn't think this would work, but now that we have cyclic imports, this works without too much friction.

I also updated all the packages I could. There were a handful that used the output from `Brioche.gitRef` directly, so I didn't think it made sense to change those (at least for now... but it might be possible if we added some extra mixin methods on the returned recipe or something?)

### DIFF
--- a/packages/amber/project.bri
+++ b/packages/amber/project.bri
@@ -1,6 +1,5 @@
 import nushell from "nushell";
 import * as std from "std";
-import { gitCheckout } from "git";
 import { cargoBuild } from "rust";
 
 export const project = {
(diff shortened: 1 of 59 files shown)


jaudiger (approved):

jaudiger:
We should open a new issue to track this work. At the end, it'll standardize all the git checkout. Once this PR is merged, we'll have two ways to do it, depending on whether we need additional git information.

PR #424, opened 2025-05-04
[std] Add `pipe` utils

kylewlacy:
This PR adds a new `std.pipe()` function, as well as an equivalent new `.pipe()` method to the `Recipe` types. For recipes, `recipe.pipe(...)` is equivalent to `std.pipe(recipe, ...)`.

We have a lot of code across many packages that look like this today:

```typescript
export default function () {
  let recipe = std.runBash`...`
    .workDir()
    .toDirectory();

  recipe = std.setEnv(recipe,  { /* env vars */ });
  recipe = std.withRunnableLink(recipe, "bin/path");

  return recipe;
}
```

(A few random examples: [lipcap](https://github.com/brioche-dev/brioche-packages/blob/2c23c69d5cfe113778214d544bb6aefbb2ae5ab6/packages/libcap/project.bri), [git](https://github.com/brioche-dev/brioche-packages/blob/2c23c69d5cfe113778214d544bb6aefbb2ae5ab6/packages/git/project.bri), [libxslt](https://github.com/brioche-dev/brioche-packages/blob/2c23c69d5cfe113778214d544bb6aefbb2ae5ab6/packages/libxslt/project.bri), [nginx](https://github.com/brioche-dev/brioche-packages/blob/2c23c69d5cfe113778214d544bb6aefbb2ae5ab6/packages/nginx/project.bri))

With the new `pipe` utils, this code will work equivalently:

```typescript
export default function () {
  return std.runBash`...`
    .workDir()
    .toDirectory()
    .pipe((recipe) => std.setEnv(recipe,  { /* env vars */ }))
    .pipe((recipe) => std.withRunnableLink(recipe, "bin/path"));
}
```

I've been fairly happy with the existing style, but it does have some annoying drawbacks. The biggest one that `pipe` fixes is no longer needing to change between `return <code>`, `const recipe = <code>` and `let recipe = <code>`.

I've had my eye on the [proposed ECMAScript pipeline operator](https://github.com/tc39/proposal-pipeline-operator), which would be a (hopefully) more ergonomic option built right into the language. But that proposal hasn't really seen much movement, and there's no guarantee at this point that it would be adopted. We _could_ implement runtime support for the operator anyway (e.g. something like [`@babel/plugin-proposal-pipeline-operator`](https://babeljs.io/docs/babel-plugin-proposal-pipeline-operator), but I'd prefer to avoid adopting syntax in Brioche that could differ from upstream ECMAScript someday. So, I felt the `pipe` method might be an interesting middle-ground.


jaudiger:
Nice work @kylewlacy 🚀

This syntax remembers me a bit the `map` in rust (https://doc.rust-lang.org/std/option/enum.Option.html#method.map, https://doc.rust-lang.org/std/iter/struct.Map.html). In Rust, it's really easy to apply a function to a container holding values. I kinda of miss this feature, when I'm developing in Java or Python, there is not always an idiomatic way to handle properly all the use cases a Rust `map` can handle. In Java, the operator exists, but it's not as advanced as in Rust.

Finally, what you did here, is really an equivalent of the [pipeline concept](https://www.nushell.sh/book/pipelines.html) of Nushell. Due to how the data is mapped in memory by Nushell, it's really easy to manipulate it through pipelines. That's the most used feature of this shell, and there is nothing similar in other shells. That's why all my DevOps scripts are written in Nushell, because I can't live without pipelines. So, you don't need to convince me here, you’re preaching to the converted 😄

I'm in favour to have this concept of pipelines for the recipe. If this PR got merged, we'll update all the packages that could take advantage of the new syntax. And maybe one day, it'll be part of ECMAScript. And way, we should stick to the standard here and we shouldn't implement a layer on top of it. That's the issue of the other package managers (and more particularly Nix), a lot of people dislike its language syntax. Here by fully following ECMAScript, we ensure that developers do not need to learn a new language or new syntaxes to write things.

jaudiger (approved):

PR #425, opened 2025-05-04
[Breaking][std] Simplify recipes

kylewlacy:
This PR makes a number of changes around the `std.Recipe` type, plus related types/methods/etc. Additionally, it replaces the `std.AsyncRecipe` type with the new `std.RecipeLike` type, which accepts either a recipe or a function that returns a recipe. Here's a list of changes:

- Replace `std.AsyncRecipe` with `std.RecipeLike`. It's an alias to accept `Awaitable<Recipe>` or a function that returns `Awaitable<Recipe>`
- Update all uses of `std.AsyncRecipe` to use `std.RecipeLike` instead
- Update all packages to avoid calling functions where possible (since `std.RecipeLike` can now take the function without needing to call it first)
- Remove `Recipe.bake()` method. I thought this method would have some use by now, but it's never been used in any package outside of `std`.
- Update `std.recipe()` to support passing a recipe-returning function. It effectively combines with the old behavior of `std.recipeFn()`.
- Add new `std.recipeFn()` function (effectively a new function now, since `std.recipe()` does what the old function did). `std.recipeFn()` wraps a function, returning a function that passes through its arguments but wraps the return value with `std.recipe()`.
- Remove internal `std.RecipeSerialization<T>` type

> **Note**: Performance is a little slower on this branch, but not significantly. On my machine, running `brioche build -p packages/aws_cli` (a relatively slow package to evaluate) went from 0.77s to 0.84s, or about 70ms slower

Here's an example of what the new `std.RecipeLike` changes enable:

```typescript
export default function (): std.Recipe<std.Directory> {
  // Previous:
  // return std.runBash`...`
  //   .dependencies(std.toolchain(), meson(), ninja(), cmake(), libcap())
  //   .workDir(source)
  //   .toDirectory();

  // New:
  return std.runBash`...`
    .dependencies(std.toolchain, meson, ninja, cmake, libcap)
    .workDir(source)
    .toDirectory();
}
```

### DIFF
--- a/packages/alsa_lib/project.bri
+++ b/packages/alsa_lib/project.bri
@@ -18,7 +18,7 @@ export default function alsaLib(): std.Recipe<std.Directory> {
     make install DESTDIR="$BRIOCHE_OUTPUT"
   `
     .workDir(source)
-    .dependencies(std.toolchain())
+    .dependencies(std.toolchain)
     .toDirectory();
 
   return std.setEnv(alsaLib, {
@@ -48,7 +48,7 @@ export async function test() {
     gcc main.c -o main -lasound
     ./main | tee "$BRIOCHE_OUTPUT"
   `
-    .dependencies(std.toolchain(), alsaLib())
+    .dependencies(std.toolchain, alsaLib)
     .env({ src: src })
     .toFile();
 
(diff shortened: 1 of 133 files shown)


jaudiger (approved):
LGTM ! Thanks @kylewlacy 🤩 The performance regression isn't that bad, I guess we can live with it as for now.

PR #446, opened 2025-05-05
refactor: use new RecipeLike when possible in dependencies and linkDependencies

jaudiger:
Follow-up of https://github.com/brioche-dev/brioche-packages/pull/425

### DIFF
--- a/packages/vdeplug4/project.bri
+++ b/packages/vdeplug4/project.bri
@@ -16,7 +16,7 @@ const source = Brioche.gitCheckout({
 export default function vdeplug4(): std.Recipe<std.Directory> {
   let vdeplug = cmakeBuild({
     source,
-    dependencies: [std.toolchain(), s2argvExecs()],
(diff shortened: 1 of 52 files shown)


packages/vdeplug4/project.bri
```diff
@@ -16,7 +16,7 @@ const source = Brioche.gitCheckout({
 export default function vdeplug4(): std.Recipe<std.Directory> {
   let vdeplug = cmakeBuild({
     source,
-    dependencies: [std.toolchain(), s2argvExecs()],
+    dependencies: [std.toolchain(), s2argvExecs],
```
kylewlacy, line 19: found one left over!

```suggestion
    dependencies: [std.toolchain, s2argvExecs],
```

kylewlacy (approved):
Looks great!

packages/vdeplug4/project.bri
jaudiger: Good catch !

PR #452, opened 2025-05-07
Quick follow up due to recent MRs

jaudiger:
Two things here:

1. Begin to make use of use std.pipe(), but it's just the beginning !
2. Properly extract the source to a const variable for ca_certificates package.


kylewlacy (approved):
:tada:

PR #462, opened 2025-05-09
feat(golangcilint): add package

jaudiger:
Add a new package [`golangci-lint`](https://github.com/golangci/golangci-lint): Fast linters runner for Go

```bash
[container@8682ebba2aec workspace]$ brioche-test packages/golangci_lint/
Build finished, completed (no new jobs) in 2.13s
Result: 796eb879fd65f900cd06a41e5c4ede85f9986f53c5853581b8acf9301910b32e
[container@8682ebba2aec workspace]$ brioche-live-update packages/golangci_lint/
Build finished, completed (no new jobs) in 5.97s
Running brioche-run
{
  "name": "golangci_lint",
  "version": "2.1.6",
  "repository": "https://github.com/golangci/golangci-lint.git",
  "extra": {
    "buildDate": "2025-05-04"
  }
}
```

### DIFF
--- a/packages/github_cli/project.bri
+++ b/packages/github_cli/project.bri
@@ -5,7 +5,9 @@ import { goBuild } from "go";
 export const project = {
   name: "github_cli",
   version: "2.72.0",
-  latestBuildDate: "2025-04-30",
+  extra: {
+    releaseDate: "2025-04-30",
+  },
 };
 
 const source = Brioche.gitCheckout({
@@ -24,7 +26,7 @@ export default function gh(): std.Recipe<std.Directory> {
         "-X",
         `github.com/cli/cli/v2/internal/build.Version=${project.version}`,
         "-X",
-        `github.com/cli/cli/v2/internal/build.Date=${project.latestBuildDate}`,
+        `github.com/cli/cli/v2/internal/build.Date=${project.extra.releaseDate}`,
       ],
     },
     path: "./cmd/gh",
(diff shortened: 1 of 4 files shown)


jaudiger:
This PR @kylewlacy contains maybe a controversial change (see [commit](https://github.com/brioche-dev/brioche-packages/pull/462/commits/287707f40cf10f8e9da02c38b0440da0b65c6746)). I tried to uniformize the build date in all the recipes that were using it. But let me know, if I should name it differently.

packages/github_cli/project.bri
```diff
@@ -5,7 +5,9 @@ import { goBuild } from "go";
 export const project = {
   name: "github_cli",
   version: "2.72.0",
-  latestBuildDate: "2025-04-30",
+  extra: {
+    buildDate: "20241220221126",
```
kylewlacy, line 9: Looks like this format differs from what the `live-update` script would write

kylewlacy (commented):
:+1: for unifying the build date fields

But... "build date" always felt a bit misleading since it's not the actual date the binary was built... but the variable for `github_cli` is called "build date" so I thought it made sense in that context specifically

So, 2 suggestions:

1. Change it to `extra.releaseDate` -- that is, it's the date this version of the project was released. (Open to other bikeshedding suggestions too)
2. Standardize the string format as ISO-8601 UTC. Using a standard format means we could even support updating this field within [`std.liveUpdateFromGithubReleases`](https://github.com/brioche-dev/brioche-packages/blob/34b51eeb7a856e207f3014e7d772fed8f588d9f0/packages/std/extra/live_update_from_github_releases.bri#L51) directly. ISO-8601 seems like an obvious choice, plus would make it easy to transform the string into a more specific format (e.g. `project.extra.releaseDate.slice(0, 10)` to get `YYYY-mm-dd`, `.slice(0, 19).replace(/[T:-]/g, "")` to get `YYYYmmddHHMMSS`, etc)

packages/github_cli/project.bri
jaudiger: I fixed it.

jaudiger:
I did the two suggestions you asked. Indeed, it's always better to align to the ISO 8601 standard.

kylewlacy (approved):
:100:

PR #481, opened 2025-05-11
feat(jpeg): add package

jaudiger:
Add a new package [`jpeg`](https://www.ijg.org): a image manipulation library for JPEG image codec.

The name of the package was also a bit hard to find (see #480). The official website uses `jpeg `, Nix uses `libjpeg_original` and homebrew simply uses `jpeg`.

I decide to choose `libjpeg` to make it clear this is the JPEG library and this is not the Turbo JPEG library. And to have a consistency with #480

### DIFF
--- a/packages/libjpeg/brioche.lock
+++ b/packages/libjpeg/brioche.lock
@@ -1,0 +1,9 @@
+{
+  "dependencies": {},
+  "downloads": {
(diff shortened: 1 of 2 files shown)


kylewlacy (approved):
LGTM

PR #488, opened 2025-05-15
Optionally authenticate requests in `std.liveUpdateFromGithubReleases`

kylewlacy:
This PR updates the Nushell script for `std.liveUpdateFromGithubReleases()` to authenticate the GitHub API request if the `$GITHUB_TOKEN` env var is set. GitHub recently announced that they were adding more restrictive rate limits for unauthenticated requests: https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/

Experimentally, the endpoint `https://api.github.com/repos/{owner}/{repo}/releases/latest` has a rate limit of 60 when unauthenticated, and resets on the hour (which can be seen with the `x-ratelimit-limit` and `x-ratelimit-reset` response headers). When using a token, the rate limit increases to 5,000 tokens (but seems to reset _after an hour_, not _on the hour_). We've got >100 packages, and we've already seen some live-update runs fail. #487 will help so failures are more evenly spread out, and this change works in tandem to reduce the number of failures-- at least when using `std.liveUpdateFromGithubReleases()`

---

Oh, and based on https://github.com/brioche-dev/brioche-packages/pull/487#issuecomment-2882494860, I also included a change to bump the Ubuntu runners to 24.04

PR #500, opened 2025-05-17
Fix typos in `std.liveUpdateFromGithubReleases()` script

kylewlacy:
Follow-up to #488

This PR fixes a typo in #488 that was causing authenticating to the GitHub API to fail (the script was adding an extra `$` to the authorization token). I also spotted another typo in the version validation while I was debugging


packages/std/extra/live_update_from_github_releases.bri
```diff
@@ -73,7 +73,7 @@ export function liveUpdateFromGithubReleases(
       }
 
       let version = $parsedTagName.0.version?
-      if version == null {
+      if $version == null {
```
jaudiger, line 76: Good catch, indeed!

jaudiger (approved):

PR #550, opened 2025-05-26
feat(scdoc): add live update capability

jaudiger:
```bash
[container@178f927a158d workspace]$ blu packages/scdoc/
 0.07s ✓ Process 946
Build finished, completed 1 job in 5.04s
Running brioche-run
{
  "name": "scdoc",
  "version": "1.11.3"
}
```

### DIFF
--- a/packages/scdoc/project.bri
+++ b/packages/scdoc/project.bri
@@ -1,4 +1,6 @@
 import * as std from "std";
+import git from "git";
+import nushell from "nushell";
 
 export const project = {
   name: "scdoc",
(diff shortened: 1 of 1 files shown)


packages/scdoc/project.bri
```diff
@@ -53,3 +55,26 @@ export async function test(): Promise<std.Recipe<std.File>> {
...
+      | from json
+      | update version $version
+      | to json
+  `);
+
+  return std.withRunnable(std.directory(), {
+    command: "nu",
+    args: [src],
+    env: { project: JSON.stringify(project) },
+    dependencies: [git, nushell],
```
kylewlacy, line 78: Is git actually needed here? Looks like it's still doing HTML-based matching

jaudiger: Indeed, it was a leftover when I was trying to get the git tags from `https://git.sr.ht/~sircmpwn/scdoc`. But it wasn't working as expected, so I decided to switch to the "normal" solution we adopted, aka retrieving the version with a Nushell script.

kylewlacy (approved):
LGTM

PR #637, opened 2025-06-08
chore(iamlive): add at least one assertion in the test

jaudiger:
Add at least one assertion in the test method of `iamlive`. Even if the package doesn't output any version, at least we should assert that it outputs something. Not ideal, but better than nothing I guess.


kylewlacy (approved):
Makes sense to me :+1:

PR #666, opened 2025-06-14
feat(lzo): add package

jaudiger:
Add a new package [`lzo`](https://www.oberhumer.com/opensource/lzo/): a portable lossless data compression library written in ANSI C

```bash
[jaudiger@lima-fedora-x86-64 brioche-packages]$ brioche run -e liveUpdate -p packages/lzo
 0.26s ✓ Process 19329
Build finished, completed 1 job in 49.55s
Running brioche-run
{
  "name": "lzo",
  "version": "2.10"
}
[jaudiger@lima-fedora-x86-64 brioche-packages]$ brioche build -e test -p packages/lzo
19379  │ 2.10
 0.22s ✓ Process 19379
Build finished, completed 1 job in 18.71s
Result: 18565e1d0ffbaff7b0dd12f6789e705c5b2c7b4395c17b623dd44dc5d9e6ec60
```

### DIFF
--- a/packages/lzo/brioche.lock
+++ b/packages/lzo/brioche.lock
@@ -1,0 +1,9 @@
+{
+  "dependencies": {},
+  "downloads": {
+    "https://www.oberhumer.com/opensource/lzo/download/lzo-2.10.tar.gz": {
+      "type": "sha256",
+      "value": "c0f892943208266f9b6543b3ae308fab6284c5c90e627931446fb49b4221a072"
+    }
+  }
+}
--- a/packages/lzo/project.bri
+++ b/packages/lzo/project.bri
@@ -1,0 +1,70 @@
+import nushell from "nushell";
+import * as std from "std";
+
+export const project = {
+  name: "lzo",
+  version: "2.10",
+};
+
+const source = Brioche.download(
+  `https://www.oberhumer.com/opensource/lzo/download/lzo-${project.version}.tar.gz`,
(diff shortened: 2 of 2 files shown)


kylewlacy:
@jaudiger Well glad to hear you got an environment working at all, but that sounds really painful to use... I'm hoping that the ARM64 packages will be ready to try out in the next day or two

jaudiger:
@kylewlacy Yesterday evening, I managed to setup a new aarch64 environment using `lima` with Rosetta enabled inside it. Meaning if a binary with its dependency is present in the VM, Rosetta will take care of the emulation. It didn't work on the first try, though. I had to use the packed mode when I installed Brioche to ensure all the `correct` dependencies were present (libc, ssl, etc). Plus, I used a modified version of the install script to select an architecture different from the host. The difference between Rosetta + Docker on MacOS is: the whole container is emulated, whereas here, only the x86_64 binaries present in the aarch64 VM are emulated. Which is kinda cool, since only a subpart of the system is not run at native speed ! This new setup mimics the [`FEX`](https://github.com/FEX-Emu/FEX) approach. And I like it more than my previous setup, plus I don't have anymore the issues with unknown backend, or segmentation fault. So, for now, I can live with this new setup. So far I'm happy with it. I'll maybe open a PR to update the install script to let the user override the architecture if he wants to use similar approach. What do you think about it ?

PR #716, opened 2025-06-23
Add support for `aarch64-linux`

kylewlacy:
This PR updates all packages plus the GitHub Actions "Build" workflow to support building for `aarch64-linux`.

To start, there's a new `std.CURRENT_PLATFORM` const which uses the new JS op from https://github.com/brioche-dev/brioche/pull/260.

`std.toolchain` also took a bit of iteration (building off some old work in https://github.com/brioche-dev/brioche-packages/commit/42b098cb474a6b1d3bbf4202f537d3b12c49fb22). But the changes there are mostly "boring", and mostly come down to using the right target triple or dynamic linker path based on platform. I updated the runtime utils to the latest commit used across `std`.

Once the toolchain was working, most packages built fine for aarch64 without any extra changes! There were a handful of exceptions though, and the following packages were tweaked:

- `bugstalker`: Doesn't support aarch64 at all today (it's marketed as an x86-64 debugger). I added an assertion to prevent building it plus a "magic comment" to skip it in the build workflow.
- `go`, `rust`, `nodejs`: Use pre-built artifacts, so needed adjustment to download the aarch64 versions of each.
- `gitui`, `xplr`: Removed upstream-provided `.cargo/config.toml`. These were overriding the aarch64 linker, which I'm guessing was set up for cross-compilation in CI? Anyway, just removing the file from the source fixed the build.
- `linux`: Required `python` as a dependency when building for aarch64 with the default configuration. Not sure why... but I added `python` as a dependency (regardless of platform) to fix it.
- `llvm`: Required more RAM to build than was available on our aarch64 runner (M1 Mac mini running Asahi Linux). As a quick hack, I just tweaked the parallelism based on platform.
- `proot`, `strace`: Had issues around 32-bit ARM personalities, so I adjusted their builds for now. We should fix this in the future.

Finally, I updated the CI/CD pipeline for aarch64 using a custom runner (M1 Mac mini with Asahi Linux). It's a lot slower than our custom x86-64 runner, so I also had to bump up the timeout for the "build" CI/CD pipeline: a full rebuild appears to take ~16 hours on aarch64 compared to ~8 hours for x86-64, and I set the new timeout to 24 hours to give some overhead.

I also updated the CI/CD pipeline to skip packages based on a "magic comment": if a package contains the string `@brioche-packages skip-platform aarch64-linux` it'll be skipped on aarch64 (also works to skip a package for x86-64). This is currently only used by `bugstalker`, since the upstream project doesn't support it (and would take a significant lift to add support, since it's a debugger and therefore deals with architecture-specific details)


packages/strace/project.bri
```diff
@@ -14,7 +14,18 @@ const source = Brioche.download(
 
 export default function strace(): std.Recipe<std.Directory> {
   return std.runBash`
-    ./configure --prefix=/
+    case "$(uname -m)" in
```
jaudiger, line 17: We can instead use `--enable-mpers=check`, see https://github.com/strace/strace/commit/3eb199a29d29c2cb44b66483d305717006b72b7f. This option let's detect if the current architecture does support multiple personalities.

packages/llvm/project.bri
```diff
@@ -22,7 +22,11 @@ export default function llvm(): std.Recipe<std.Directory> {
       CMAKE_BUILD_TYPE: "Release",
     },
     env: {
-      CMAKE_BUILD_PARALLEL_LEVEL: "16",
+      // HACK: Building LLVM can use A LOT of memory, especially when building
+      // in parallel. These values are tuned based on the CI runners currently
+      // in use in the `brioche-packages` repo.
+      CMAKE_BUILD_PARALLEL_LEVEL:
```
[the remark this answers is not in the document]
kylewlacy: I know LLVM's build is heavily parallelizable, so I'd expect going from 16 to 8 would double the build time, meaning it'd go from about 1 hour to roughly 2 hours. I did queue up a test build here to confirm: https://github.com/brioche-dev/brioche-packages/actions/runs/15839816444

If that is the case, then I'd rather keep it set per architecture like this as a temporary measure. I think this code is worth trimming an hour off the build, at least for now. Higher level, this is related to https://github.com/brioche-dev/brioche-packages/discussions/230: we'll probably want to have some way to set the build parallelism baesd on which runner its on or something

packages/strace/project.bri
jaudiger: Might be a toolchain issue or maybe it's simply a limitation on this architecture. I wasn't't able to find any documentation on it. Except that in Nixpkgs, the flag is always set: https://github.com/NixOS/nixpkgs/blob/c7ab75210cb8cb16ddd8f290755d9558edde7ee1/pkgs/by-name/st/strace/package.nix#L42 and on Homebrew, they simply disabled it (but that's not the right option: https://github.com/Homebrew/homebrew-core/blob/29857a21b7560c2e0bb826b183cfaa8b0c1871fe/Formula/s/strace.rb#L24

Sure, we can leave it as is, and see later by opening an issue.

PR #717, opened 2025-06-25
NPM packages live update

jaudiger:
This PR refactors multiple project files to improve maintainability and standardize the implementation of live updates for NPM packages. It introduces a new utility, `std.liveUpdateFromNpmPackages`, replacing custom implementations of `liveUpdate` across various packages. Additionally, it consolidates version-matching logic into a shared constant and improves documentation for related utilities.

I took this opportunity to move all the expected live-update fields of a NPM package to extra:

```typescript
export const project = {
  name: "aws_cdk",
  version: "2.1019.1",
  extra: {
    packageName: "aws-cdk",
  },
};
```

This is a bit different than the GitHub live update approach where the live-update fields are directly at the project root. But it makes sense here since we're talking about a repository URL.

Another difference between `std.liveUpdateFromGithubReleases` and `std.liveUpdateFromNpmPackages` comes from the options interface:

```typescript
interface LiveUpdateFromGithubReleasesOptions {
  project: { version: string; repository: string };
  matchTag?: RegExp;
}

...

interface LiveUpdateFromNpmPackagesOptions {
  project: { version: string; extra: LiveUpdateFromNpmPackagesExtraOptions };
  matchVersion?: RegExp;
}
```

We have on one side a `matchTag` and on the other side a `matchVersion`, at the end we are talking about version. even when we look at the GitHub `version/tag` extraction we are using both terminologies:

```nu
      # Include GitHub Token if present (for increased rate limits)
      mut gh_headers = []
      if ($env.GITHUB_TOKEN? | default "") != "" {
        $gh_headers ++= [Authorization $'Bearer ($env.GITHUB_TOKEN)']
      }

      let tagName = http get --headers $gh_headers $'https://api.github.com/repos/($env.repoOwner)/($env.repoName)/releases/latest'
        | get tag_name

      let parsedTagName = $tagName | parse --regex $env.matchTag
      if ($parsedTagName | length) == 0 {
        error make { msg: $'Latest release tag ($tagName) did not match regex ($env.matchTag)' }
      }

      let version = $parsedTagName.0.version?
      if $version == null {
        error make { msg: $'Regex ($env.matchTag) did not include version when matching latest release tag ($tagName)' }
      }

      $env.project
        | from json
        | update version $version
        | to json
```

So, I also updated the name of the constant regex to be `DEFAULT_LIVE_UDPATE_REGEX_VERSION_MATCH`, but I didn't update the interface `LiveUpdateFromGithubReleasesOptions`, what do you think @kylewlacy should we uniformize all of that ?

### DIFF
--- a/packages/aws_cdk/project.bri
+++ b/packages/aws_cdk/project.bri
@@ -1,16 +1,17 @@
 import * as std from "std";
 import { npmInstallGlobal } from "nodejs";
-import nushell from "nushell";
 
 export const project = {
   name: "aws_cdk",
   version: "2.1019.1",
-  packageName: "aws-cdk",
+  extra: {
+    packageName: "aws-cdk",
+  },
 };
 
 export default function awsCdk(): std.Recipe<std.Directory> {
   return npmInstallGlobal({
-    packageName: project.packageName,
+    packageName: project.extra.packageName,
     version: project.version,
   }).pipe((recipe) => std.withRunnableLink(recipe, "bin/cdk"));
 }
@@ -34,24 +35,6 @@ export async function test(): Promise<std.Recipe<std.File>> {
   return script;
 }
 
-export function liveUpdate(): std.WithRunnable {
-  const src = std.file(std.indoc`
-    let releaseData = http get https://registry.npmjs.org/${project.packageName}/latest
-
-    let version = $releaseData
-      | get version
-      | str replace --regex '^v' ''
-
-    $env.project
-      | from json
-      | update version $version
-      | to json
-  `);
-
-  return std.withRunnable(std.directory(), {
-    command: "nu",
-    args: [src],
-    env: { project: JSON.stringify(project) },
-    dependencies: [nushell],
-  });
+export function liveUpdate(): std.Recipe<std.Directory> {
+  return std.liveUpdateFromNpmPackages({ project });
 }
--- a/packages/bash_language_server/project.bri
+++ b/packages/bash_language_server/project.bri
@@ -1,16 +1,17 @@
 import * as std from "std";
 import { npmInstallGlobal } from "nodejs";
-import nushell from "nushell";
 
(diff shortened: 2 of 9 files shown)


jaudiger:
To be merged after https://github.com/brioche-dev/brioche-packages/pull/716

packages/std/extra/live_update_from_npm_packages.bri
```diff
@@ -0,0 +1,144 @@
...
+ * @param project - The project export that should be updated. Must include a
+ *   `extra.packageName` property containing the name of the NPM package.
+ * @param matchVersion - A regex value (`/.../`) to extract the version number from
+ *   a string. The regex must include a group named "version". If not
+ *   provided, an optional "v" prefix will be stripped and the rest of the
+ *   version will be checked if it's a semver or semver-like version number.
+ */
+interface LiveUpdateFromNpmPackagesOptions {
+  project: { version: string; extra: LiveUpdateFromNpmPackagesExtraOptions };
+  matchVersion?: RegExp;
```
kylewlacy, line 30: ...any idea how useful this option will be? I don't think I've ever seen an NPM package that didn't use an `x.y.z` style version number (unlike git tags, where we already have several packages with weird tag names). I also haven't looked into NPM's rules to see what they allow for versions, I might try and do some research on this as a follow-up....

kylewlacy (approved):
LGTM!

Left some notes that might be good to do as a follow-up. I'm happy to merge as-is since 1) it makes the code better now and 2) I like the idea of testing it end-to-end during the next batch of live updates

packages/std/extra/live_update_from_npm_packages.bri
jaudiger: There is no need indeed to have the `matchVersion` field here if SemVer is respected.

PR #753, opened 2025-07-01
Add GitLab live update helper method

jaudiger:
Similar to `std.liveUpdateFromGithubReleases()`, this PR introduces a new helper method for GitLab forge. The options are quite similar, and only the URL check is different at the end.

Since GitLab can be deployed on premise, there is a limitation right now with the new method `std.liveUpdateFromGitlabReleases()`, we cannot override the GitLab instance URL. If this PR is merged as is. I'll open an "good first issue" to later implement this possibility.

### DIFF
--- a/packages/std/extra/live_update_from_github_releases.bri
+++ b/packages/std/extra/live_update_from_github_releases.bri
@@ -24,7 +24,7 @@ interface LiveUpdateFromGithubReleasesOptions {
 /**
  * Return a runnable recipe to live-update a project based on the latest release
  * tag from a GitHub repository. The project's version will be set based on a
- * regex match against the latest tag name. The repository is inferrred from the
+ * regex match against the latest tag name. The repository is inferred from the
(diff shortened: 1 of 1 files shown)


jaudiger:
Hum, I think the copilot review went from my GitHub Copilot subscription, I'll disable it next time.

jaudiger:
@kylewlacy Here is a quick testing with a new package that is not yet upstreamed:

```typescript
export function liveUpdate(): std.Recipe<std.Directory> {
  return std.liveUpdateFromGitlabReleases({ project });
}
```

```bash
Build finished, completed 1 job in 8.10s
Running brioche-run
{
  "name": "libtiff",
  "version": "4.7.0",
  "repository": "https://gitlab.com/libtiff/libtiff.git"
}

⏵ Task `Run package live-update` finished successfully
``

kylewlacy (approved):
:+1:

PR #781, opened 2025-07-04
refactor(std): move live update methods into a new folder

jaudiger:
This PR refactors a bit the live update feature that lives in the `std` package. First all the live update methods are moved inside a new folder to better segment the files hierarchy and to not raise the number of files inside one `std.extra` directory. Second, all the Nushell scripts have been extracted into normal Nu files. The advantages of that is: we can now use the usual Nu LSP to work on Nu scripts and we can even run the Nu scripts as standalone scripts to debug them if needed (of course a preliminary environment would need to be setup to at least set all the environment variables the scripts required).

I wanted to rework that before updating and adding new live update methods.

### DIFF
--- a/packages/std/extra/index.bri
+++ b/packages/std/extra/index.bri
@@ -6,10 +6,7 @@ export * from "./bash_runnable.bri";
 export * from "./set_env.bri";
 export * from "./with_runnable_link.bri";
 export * from "./apply_patch.bri";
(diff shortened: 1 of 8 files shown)


kylewlacy (approved):
LGTM, breaking these scripts out into their own files will definitely make it easier to work on them too :+1:

PR #785, opened 2025-07-06
Update automatically GitHub release date

jaudiger:
This PR enhances the GitHub release live update capabilities to also update the release date when the package relies on it in its recipe. With this PR, 3 packages are still using a custom GitHub release implementation:

- curl
- expat
- icu

Bu the good news is we want from 8 custom implementations to now 3. Of course, I'm not counting here the custom implementations we have for GitHub tags, these implementations will be refactored later with a new live update method.

Talking back of GitHub release, all the remaining packages using a custom implementation relies on specific version representation (dash, hyphen, underscore). We could, in a future work, add it to the std GitHub release live update.

### DIFF
--- a/packages/std/extra/live_update/scripts/live_update_from_npm_packages.nu
+++ b/packages/std/extra/live_update/scripts/live_update_from_npm_packages.nu
@@ -1,4 +1,12 @@
-let version = http get $'https://registry.npmjs.org/($env.packageName)/latest'
+# Get project metadata
+mut project = $env.project
+  | from json
+
+# Retrieve the latest release information from npm registry
+let releaseInfo = http get $'https://registry.npmjs.org/($env.packageName)/latest'
+
+# Extract the version
+let version = $releaseInfo
   | get version
 
 let parsedVersion = $version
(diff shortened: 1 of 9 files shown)


packages/std/extra/live_update/scripts/live_update_from_npm_packages.nu
```diff
@@ -1,4 +1,12 @@
-let version = http get $'https://registry.npmjs.org/($env.packageName)/latest'
+# Get project metadata
+mut project = $env.project
+  | from json
+
+# Retrieve the latest release information from npm registry
```
jaudiger, line 5: The naming `npm` respects the one of this [PR](https://github.com/brioche-dev/brioche-packages/pull/783)

jaudiger:
> Talking back of GitHub release, all the remaining packages using a custom implementation relies on specific version representation (dash, hyphen, underscore). We could, in a future work, add it to the std GitHub release live update.

I got a few ideas on how to do it. I'll start to work on it, once this work is upstream.

PR #786, opened 2025-07-06
chore(live-update): better handle when the field `repo` is not present

jaudiger:
~~Improve the stack trace when the field `repo` is missing and the live update method is executed.~~ Improve the detection when the field `repo` is missing by ensuring to detect it during static checking. The [previous behavior](https://github.com/brioche-dev/brioche-packages/pull/786/commits/bfb06bdc7b96bccd76f0216649217d4a1c0b5681) which was relying on the nullability check wasn't that easy to implement without modifying the way the constant variable `project` is cast to `LiveUpdateFromGithubReleasesOptions` or `LiveUpdateFromGitlabReleasesOptions`.

Before:

```bash
Error: error when calling liveUpdate

Caused by:
    TypeError: Cannot read properties of undefined (reading 'match')
        at tryParseGithubRepo (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/std/extra/live_update/from_github_releases.bri:84:22)
        at parseGithubRepo (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/std/extra/live_update/from_github_releases.bri:111:16)
        at Module.liveUpdateFromGithubReleases (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/std/extra/live_update/from_github_releases.bri:53:35)
        at Module.liveUpdate (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/curl/project.bri:96:14)
```

After (same stack trace), but the checker this time detected the issue:

```bash
[Error] file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/curl/project.bri:98:5
Property 'repository' is missing in type '{ name: string; version: string; }' but required in type '{ version: string; readonly repository: string; }'.
```

### DIFF
--- a/packages/std/extra/live_update/from_github_releases.bri
+++ b/packages/std/extra/live_update/from_github_releases.bri
@@ -17,8 +17,8 @@ import type {} from "nushell";
  *   tag will be checked if it's a semver or semver-like version number.
  */
 interface LiveUpdateFromGithubReleasesOptions {
-  project: { version: string; repository: string };
-  matchTag?: RegExp;
+  project: { version: string; readonly repository: string };
+  readonly matchTag?: RegExp;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


kylewlacy (approved):
LGTM :+1:

I didn't expect `readonly` would change the error message here, interesting! Are there more places that we should be using `readonly` then? (Or maybe there's some way for us to improve the error message for non-readonly fields?)

PR #791, opened 2025-07-07
Add live update version normalizer

jaudiger:
Add a version normalizer for GitHub and GitLab live update releases. This new capability will be used with the following packages:

- curl (this PR)
- expat (will be done in an upcoming PR)
- icu (will be done in an upcoming PR)

`expat` and `icu` require project version formatter to convert `X.Y.Z` to either `X-Y-Z` or `X_Y_Z`. I'm planning to add a new std helper to convert a package version. Why this approach instead of simply using a one line function in each package ?

1. Factorize the code inside a common package (here std)
2. Later we could add function to extract sub part of a version (this is sometimes needed in testing package function, ie: `libcap`)

### DIFF
--- a/packages/curl/project.bri
+++ b/packages/curl/project.bri
@@ -2,12 +2,12 @@ import * as std from "std";
 import brotli from "brotli";
 import libpsl from "libpsl";
 import libssh2 from "libssh2";
-import nushell from "nushell";
 import openssl from "openssl";
 
 export const project = {
   name: "curl",
   version: "8.14.1",
+  repository: "https://github.com/curl/curl.git",
 };
 
 const source = Brioche.download(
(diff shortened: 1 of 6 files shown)


packages/std/extra/live_update/from_github_releases.bri
```diff
@@ -72,6 +77,7 @@ export function liveUpdateFromGithubReleases(
         repoOwner,
         repoName,
         matchTag: matchTag.source,
+        normalizeVersion: normalizeVersion.toString(),
```
jaudiger, line 80: This conversion is expected. Environment variables need to always be a string.

packages/curl/project.bri
```diff
@@ -2,12 +2,12 @@ import * as std from "std";
...
-import nushell from "nushell";
 import openssl from "openssl";
 
 export const project = {
   name: "curl",
   version: "8.14.1",
+  repository: "https://github.com/curl/curl.git",
 };
 
 const source = Brioche.download(
```
jaudiger, line 13: To see if later we should update the source of truth for this package since we now have the field `repository` in the `project`.

kylewlacy (approved):
Ah, great little tweak!

PR #807, opened 2025-07-09
Project version formatting

jaudiger:
Finally finish what was began in https://github.com/brioche-dev/brioche-packages/pull/785 and https://github.com/brioche-dev/brioche-packages/pull/791.

This PR removes the remaining GitHub release custom live update implementation (found in `icu` and `expat)`. `https://api.github.com/repos/.../releases/latest` can now only be found in the Nushell script of the GitHub release live update. Meaning all the API call for checking the latest release available make use of the GitHub token (at least in the CI), reducing the API throttling, etc.

Also part of this PR, there are new methods in the std package to manipulate the project version:

- `std.projectVersionToDashFormat`
- `std.projectVersionToUnderscoreFormat`

As the name implies, they convert the project version to a specific format. All the packages relying on the dash and underscore formats have been updated.

### DIFF
--- a/packages/std/extra/project_version/format.bri
+++ b/packages/std/extra/project_version/format.bri
@@ -1,0 +1,139 @@
+import { PROJECT_VERSION_SEMVER_MATCH } from "./index.bri";
+
+/**
+ * All the possible project version format type.
+ */
+enum ProjectVersionFormatType {
+  DOT = ".",
+  DASH = "-",
+  UNDERSCORE = "_",
+}
+
+/**
+ * Options for the project version format.
+ */
+export interface ProjectVersionFormatOptions {
+  project: {
+    version: string;
+  };
+}
+
+/**
+ * Formats the project version to a dash format.
+ *
+ * @param options - Options for the project version format.
+ *
+ * @returns The formatted project version.
+ *
+ * @throws Error if the project version is not a semver-like version.
+ *
+ * @example
+ * ```typescript
+ * export const project = {
+ *   name: "brioche",
+ *   version: "0.1.0",
+ *   extra: {
+ *     versionDash: "0-1-0",
+ *   },
+ * };
+ *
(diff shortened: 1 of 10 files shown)


packages/std/extra/project_version/format.bri
```diff
@@ -0,0 +1,139 @@
+import { PROJECT_VERSION_SEMVER_MATCH } from "./index.bri";
+
+/**
+ * All the possible project version format type.
+ */
+enum ProjectVersionFormatType {
+  DOT = ".",
+  DASH = "-",
+  UNDERSCORE = "_",
+}
```
kylewlacy, line 10: #nonblocking

At some point, I think I want to enable the [`--erasableSyntaxOnly`](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8/#the---erasablesyntaxonly-option) TypeScript option, so I'd prefer to avoid enums when possible

(I also just generally think enums have a few footguns compared to other ways to write the same thing, so I've avoided them for a long time too)

```suggestion
/**
 * All the possible project version format type.
 */
const ProjectVersionFormatType = {
  DOT: ".",
  DASH: "-",
  UNDERSCORE: "_",
} as const;
```

This is a minor nitpick though and it would be easy to tweak in the future so I'm fine with leaving it as a future change too

kylewlacy (approved):
LGTM, had one piece of feedback around the use of `enum` but also not a blocker for merging

packages/std/extra/project_version/format.bri
jaudiger: I'll do it, no problem. I didn't know about the [--erasableSyntaxOnly](https://devblogs.microsoft.com/typescript/announcing-typescript-5-8/#the---erasablesyntaxonly-option) TypeScript option. Indeed, let's make the activation of this option less painful by preparing the code for it 😅

PR #833, opened 2025-07-12
refactor(git): use std.process() instead of std.runBash() for gitCheckout()

jaudiger:
Since the git invocation commands was done through a single call, we can now switch to `std.process()` instead of using `std.runBash()`. It better enforces the process isolation, I think.


kylewlacy (approved):
Nice!

PR #834, opened 2025-07-13
feat(rust): generate `Cargo.lock` when the file does not exist in the upstream source

jaudiger:
Sometimes the upstream source of a Rust package does not contain any `Cargo.lock` file which is not recommended, since we cannot ensure build reproducibility. I think, we should try generating the lock file when we encounter such rust packages. It's more a workaround, and usually this issue should be fixed upstream.

Here is a list of packages for example that do not contain any lock files:

- https://github.com/taiki-e/cargo-hack
- https://github.com/taiki-e/cargo-llvm-cov
- https://github.com/taiki-e/cargo-minimal-versions
- https://github.com/taiki-e/cargo-no-dev-deps

### DIFF
--- a/packages/rust/project.bri
+++ b/packages/rust/project.bri
@@ -187,7 +187,7 @@ interface CargoBuildParameters {
  *   unnecessarily re-downloading dependencies when the source changes.
  *   Defaults to `true` (should only be disabled when there's an upstream
  *   issue with `cargo chef`).
- * @param generateLockfile - Controls if the crate from `source` should
+ * @param unsafeGenerateLockfile - Controls if the crate from `source` should
  *   generate a `Cargo.lock` file before the build. Defaults to `false`.
  * @param buildParams - Optional build parameters.
  */
@@ -199,7 +199,7 @@ export interface CargoBuildOptions {
   env?: Record<string, std.ProcessTemplateLike>;
   unsafe?: std.ProcessUnsafeOptions;
   cargoChefPrepare?: boolean;
-  generateLockfile?: boolean;
+  unsafeGenerateLockfile?: boolean;
   buildParams?: CargoBuildParameters;
 }
 
@@ -235,7 +235,7 @@ export function cargoBuild(options: CargoBuildOptions) {
   // Vendor the crate's dependencies
   const crate = vendorCrate({
     source: options.source,
-    generateLockfile: options.generateLockfile,
+    unsafeGenerateLockfile: options.unsafeGenerateLockfile,
     cargoChefPrepare: options.cargoChefPrepare,
   });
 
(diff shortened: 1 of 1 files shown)


jaudiger:
After further investigation, I found something that could be interested, and maybe we should not merge this PR as is.

1. First, if we want to merge this PR maybe we could switch from a `allow` pattern to a `deny` pattern. I mean, by default, we shouldn't try to generate the `Cargo.lock` unless explicitly specified by the user. To do that, we could introduce a new field in the interface `VendorCrateOptions`:

```typescript
interface VendorCrateOptions {
  source: std.RecipeLike<std.Directory>;
  generateLockFile?: boolean; // The name could be subject to change
  cargoChefPrepare?: boolean;
}
```

This way, we could track the Rust crates that do not have a `Cargo.lock`. And if `generateLockFile` is true, and a lock file is already present, we exit in error to track crates that could later add the lock file.

2. Following a discussion [here](https://github.com/taiki-e/cargo-llvm-cov/pull/152), there is another way to download the crate's source. [crates.io](https://crates.io) exposes an API dedicated to the downloading of the source code, and it seems that crates published on this registry do always contain the `Cargo.lock` file.

```bash
> curl -L https://crates.io/api/v1/crates/cargo-hack/0.6.37/download | tar -xz
> ls cargo-hack-0.6.37/
src   tests   Cargo.lock   Cargo.toml   Cargo.toml.orig   CHANGELOG.md   LICENSE-APACHE   LICENSE-MIT  󰂺 README.md
```

So, we could do, in addition to the first point, introduce a new method defined in the Brioche `rust` package to download a crate source from the [crates.io](https://crates.io) registry. And we should then switch to this default download method for all the Rust crates published on this registry. Since, this registry should be the source of truth for the Rust ecosystem. We could even further expand that by introducing a new std helper to check directly on the registry the latest version available for a crate. As we are already doing for npm packages.

jaudiger:
Ok, I'll use this branch to work on the points mentioned in my comment then.

> (I also really want to have a guide for package contributors sooner rather than later... keeping notes for issues like this in one place will be helpful over time!)

Agreed, at some point (soon), we should have a guide to help the packaging on Brioche. If in August, I finished the work on the std helper for version testing, I could then switch on it.

kylewlacy (approved):
LGTM

Happy with how this looks overall! Had two minor nitpicks around the public interface though (but fine getting this in now and saving some cleanup/iteration as a follow-up):

1. `lockFile` -> `lockfile` - Cargo _usually_ uses the term "lockfile" as one word rather than "lock file"
2. Since this function is ["unsafe"](https://brioche.dev/docs/how-it-works/sandboxing/#rules-for-unsafe) (by Brioche's definition), I'd kinda like a way to make sure callers have to use the term "unsafe" somewhere when calling it-- similar to Rust, I think it's nice just doing Ctrl-F for "unsafe" finds potential suspicious places in a project. Maybe if we made the param something like `options: { unsafeGenerateLockfile?: boolean }` or similar?

kylewlacy:
Awesome, looks great!

PR #844, opened 2025-07-16
feat(helix): 📦️ add package

jokeyrhyme:
add a package for https://helix-editor.com/

### DIFF
--- a/packages/helix/brioche.lock
+++ b/packages/helix/brioche.lock
@@ -1,0 +1,8 @@
+{
+  "dependencies": {},
+  "git_refs": {
+    "https://github.com/helix-editor/helix.git": {
+      "25.07.1": "a05c151bb6e8e9c65ec390b0ae2afe7a5efd619b"
+    }
+  }
+}
--- a/packages/helix/project.bri
+++ b/packages/helix/project.bri
@@ -43,11 +43,19 @@ export default function helix(): std.Recipe<std.Directory> {
 }
 
 export async function test(): Promise<std.Recipe<std.File>> {
-  return nushellRunnable`
-    use std/assert
-    assert (((^hx --version | complete).stdout | str trim) == 'helix ${project.version}')
+  const script = std.runBash`
+    hx --version | tee "$BRIOCHE_OUTPUT"
   `
-    .dependencies(helix);
+    .dependencies(helix)
+    .toFile();
+
+  const result = (await script.read()).trim();
+
+  // Check that the result contains the expected version
+  const expected = `helix ${project.version}`;
+  std.assert(result === expected, `expected '${expected}', got '${result}'`);
+
+  return script;
 }
 
 export async function liveUpdate(): Promise<std.Recipe<std.Directory>> {
--- a/packages/helix/project.bri
+++ b/packages/helix/project.bri
@@ -1,0 +1,58 @@
+import * as std from "std";
+import git from "git";
+import { nushellRunnable } from "nushell";
+import rust, { cargoBuild } from "rust";
+
+export const project = {
+  name: "helix",
+  version: "25.07.1",
+  repository: "https://github.com/helix-editor/helix.git",
+};
+
+const source = Brioche.gitCheckout({
+  repository: project.repository,
+  ref: project.version,
+});
+
+export default function helix(): std.Recipe<std.Directory> {
+  // TODO: https://docs.helix-editor.com/building-from-source.html#note-to-packagers
+  // - set HELIX_DEFAULT_RUNTIME to $BRIOCHE_OUTPUT/lib/helix/runtime
+  // TODO: https://docs.helix-editor.com/building-from-source.html#configure-the-desktop-shortcut
+  // - setup desktop shortcut and icon
+  // TODO: parse languages.toml with `nu`, `std.download()` grammar sources
+  return std.merge(
+    cargoBuild({
+      env: {
+        HELIX_DISABLE_AUTO_GRAMMAR_BUILD: "1",
+      },
+      path: "helix-term",
+      runnable: "bin/hx",
+      source,
+    }),
+    std.directory({
+      lib: std.directory({
+        helix: std.directory({
+          runtime: std.directory({
+            queries: source.get('runtime/queries'),
+            themes: source.get('runtime/themes'),
+          }),
+        }),
+      }),
+    }),
+  );
+}
+
+export async function test(): Promise<std.Recipe<std.File>> {
+  return nushellRunnable`
+    use std/assert
+    assert (((^hx --version | complete).stdout | str trim) == 'helix ${project.version}')
+  `
+    .dependencies(helix);
+}
+
+export async function liveUpdate(): Promise<std.Recipe<std.Directory>> {
+  return std.liveUpdateFromGithubReleases({
+    matchTag: /^(?<version>\d+\.\d+(?:\.\d+)?)$/,
+    project,
+  });
+}
--- a/packages/helix/project.bri
+++ b/packages/helix/project.bri
@@ -1,7 +1,5 @@
 import * as std from "std";
-import git from "git";
-import { nushellRunnable } from "nushell";
-import rust, { cargoBuild } from "rust";
+import { cargoBuild } from "rust";
 
 export const project = {
   name: "helix",
@@ -33,8 +31,8 @@ export default function helix(): std.Recipe<std.Directory> {
       lib: std.directory({
         helix: std.directory({
           runtime: std.directory({
-            queries: source.get('runtime/queries'),
-            themes: source.get('runtime/themes'),
+            queries: source.get("runtime/queries"),
+            themes: source.get("runtime/themes"),
           }),
         }),
       }),


packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
+import * as std from "std";
+import git from "git";
+import rust, { cargoBuild } from "rust";
+
+export const project = {
+  name: "helix",
+  version: "25.01.1",
```
jaudiger, line 7: Helix version `25.07` has been released yesterday: https://github.com/helix-editor/helix/releases/tag/25.07 🚀

packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
...
+    runnable: "hx",
+    source,
+  });
+}
+
+export async function test(): Promise<std.Recipe<std.File>> {
+  const script = std.runBash`
+    helix --version | tee "$BRIOCHE_OUTPUT"
+  `
+    .dependencies(rust, helix)
```
jaudiger, line 31: Once the package built, to test it, you don't need the `rust` package:

```suggestion
    .dependencies(helix)
```

Thus, after that, you could simply remove `rust` from the imports

packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
...
+    },
+    path: "helix-term",
+    runnable: "hx",
+    source,
+  });
+}
+
+export async function test(): Promise<std.Recipe<std.File>> {
+  const script = std.runBash`
+    helix --version | tee "$BRIOCHE_OUTPUT"
```
jaudiger, line 29: The name of Helix binary is `hx` and not `helix`:

```suggestion
    hx --version | tee "$BRIOCHE_OUTPUT"
```

packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
...
+  ref: project.version,
+});
+
+export default function helix(): std.Recipe<std.Directory> {
+  return cargoBuild({
+    env: {
+      HELIX_DISABLE_AUTO_GRAMMAR_BUILD: "1",
+    },
+    path: "helix-term",
+    runnable: "hx",
```
jaudiger, line 22: To see the actual output of the package built, you can run `brioche build -o /tmp/output -p packages/helix`, and in the folder output, you can see:

```bash
> ls /tmp/output
bin  brioche-resources.d  brioche-run
> ls /tmp/output/bin
hx
```

```suggestion
    runnable: "bin/hx",
```

packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
...
+    source,
+  });
```
jaudiger, line 24: There is a way to build the tree sitter grammar, you could remove your env var `HELIX_DISABLE_AUTO_GRAMMAR_BUILD`, and include `git` in the dependencies + enabling network access during the building. This way the grammars will be pulled, and built:


```suggestion
  return cargoBuild({
    source,
    path: "helix-term",
    dependencies: [git],
    runnable: "bin/hx",
    unsafe: { networking: true },
  });
```

packages/helix/project.bri
```diff
@@ -0,0 +1,48 @@
...
+  const script = std.runBash`
+    helix --version | tee "$BRIOCHE_OUTPUT"
+  `
+    .dependencies(rust, helix)
+    .toFile();
+
+  const result = (await script.read()).trim();
+
+  // Check that the result contains the expected version
+  const expected = `helix ${project.version}`;
```
jaudiger, line 37: I saw that `hx --version` outputs: `helix 25.01.1 (e7ac2fcd)`, if you want to do an exact match with the short git commit hash, you could look at the package `proot`, we are using a trick there to check the short git commit hash in the test method.

jaudiger (commented):
Thank you @jokeyrhyme for upstreaming `helix` as a Brioche package  🚀 I put a few comments below, after that I think we could merge it !

packages/helix/brioche.lock
```diff
@@ -0,0 +1,12 @@
+{
+  "dependencies": {
```
jaudiger, line 2: The dependencies here should be empty: `"dependencies": {}`. Since we are in a Brioche workspace [here](https://github.com/brioche-dev/brioche-packages/blob/main/brioche_workspace.toml), we can directly make reference to other packages. I think, if you re-run a Brioche command to build the package or run the test method, Brioche will automatically update the lock file and suppress the dependencies.

kylewlacy: Ooh, interesting! @jokeyrhyme did you originally have `helix` in a separate dir then moved it into `brioche-pacakages`? If so then I see how the lockfile ended up like this, otherwise I'm not sure how the lockfile would've ended up like this...

jokeyrhyme: yeah, i developed these elsewhere including running `brioche install` on them, and then moved them to their current locations for the PR

you can relax, haha :)

packages/helix/project.bri
jokeyrhyme: yeah, i wasn't sure whether we were leaning towards 100% hermetic builds

there's probably a way to seal the files required for the grammar to build ...

jaudiger: I first thought that retrieving the package source from a GitHub tarball as pointed it out [here](https://github.com/NixOS/nixpkgs/blob/32a4e87942101f1c9f9865e04dc3ddb175f5f32e/pkgs/by-name/he/helix/package.nix#L15) would be enough, but when I quickly tried I wasn't able to build Helix entirely. Maybe I misconfigured something, I haven't further dig.

jokeyrhyme:
okay, bumped helix's version again, and fixed the `test()` and `liveUpdate()` exports

i initially started with a copy-paste of the `test()` export from the "bat" package, but couldn't get it to pass, but it turns out that's an issue with "bat" and probably many other packages using similar `std.runBash` recipes

```
 ~/G/b/brioche-packages > brioche run --export test --keep-temps --project ./packages/bat
Build finished, completed (no new jobs) in 1.10s
Error: artifact returned a file, expected a directory
```

kylewlacy (approved):
LGTM! It seems like the basic editor is working now (AFAICT), so I'm fine with merging as-is and iterating on `helix` via follow-up PRs

@jokeyrhyme before I hit merge, were there any of the TODOs you wanted to get in before this gets merged as an initial version?

packages/helix/project.bri
kylewlacy: Personally, I am not super picky about perfect hermeticity! I haven't looked super closely at how Helix is downloading languages, but first impressions is it looks like it's pretty stable? If so, it seems like we don't lose much by just enabling an unsafe build (we do lose caching if any of the upstream URLs are unavailable), and just using unsafe would make it more maintainable, since we don't need to deal with their config file manually. Then again, it seems like it's not a _lot_ of work to parse and download?

...anyway, I'm pretty split either way, I think either just allowing unsafe downloads or manually parsing would be fine

jokeyrhyme:
> @jokeyrhyme before I hit merge, were there any of the TODOs you wanted to get in before this gets merged as an initial version?

@kylewlacy i'm not sure what the comment conventions are in this project, but i usually go with FIXME=blocker and TODO=later 🤷 

i'm still mucking about with trying to get this perfectly hermetic, haha

the languages.toml file in upstream source statically defines the additional remotes, but i'm finding it tricky reading and parsing this out early enough to be fed into `Brioche.gitCheckout()`

and if i TODO that, then the next tricky part is that `hx --grammar build` outputs to $HOME/.config/helix/runtime by default, so i'm still learning how to preserve/move from $HOME (in the std.process sandbox) to the final $BRIOCHE_OUTPUT/lib/helix/runtime

kylewlacy:
> i initially started with a copy-paste of the `test()` export from the "bat" package, but couldn't get it to pass, but it turns out that's an issue with "bat" and probably many other packages using similar `std.runBash` recipes
> 
> ```
>  ~/G/b/brioche-packages > brioche run --export test --keep-temps --project ./packages/bat
> Build finished, completed (no new jobs) in 1.10s
> Error: artifact returned a file, expected a directory
> ```

Ah, I just caught this as well. "Tests" are pretty loosely-defined today, but I think what you were seeing for `bat` is basically expected for how tests are structured. Today, the CI pipeline basically just runs `brioche build -e test`, and the tests passes when that commands succeeds (i.e. it just "builds" the the test without running it). Generally, the test just outputs a file on success, but we basically ignore it's output... it's a restriction of `brioche build` today that the function has to return _some_ kind of artifact as a result

As for _why_ we just build tests instead of running them: (1) `build` only has access to the sandbox, whereas `run` runs on the host system, so it's more isolated, and (2) the test result is cached, so we could in theory have more expensive tests in CI and they would only need to re-run if their inputs change, just like any build)

We've had an issue for adding a `brioche test` subcommand for quite a while (https://github.com/brioche-dev/brioche/issues/94), which would hopefully help make this clearer! (Or at minimum, this is another piece that could be documented in a contributors' guide!)

jokeyrhyme:
thanks for the hints @kylewlacy 

the other approach that occurred to me was abusing `liveUpdate()` so that it doesn't just return the latest GitHub Release version, but also does something with languages.toml so that we're getting the "version" of each tree sitter grammar at the same time

still not sure where i'd go from there

i'll muck about with the code snippets you've shared and see how they go

PR #886, opened 2025-07-20
feat(live-update): better handle GitHub API errors in `std.liveUpdateFromGithubReleases()`

jaudiger:
Better handle the most common GitHub API errors. It'll be useful when we'll improve `std.liveUpdateFromGithubReleases()` to retrieve the last X releases, since it could involve using more and more this GitHub helper. Plus, we could make use of the same trick when introducing the `std.liveUpdateFromGithubTags()` (see https://github.com/brioche-dev/brioche-packages/pull/852#pullrequestreview-3035881259). And for new contributors, these GitHub helpers will be more comprehensive in case a "normal" error happens, see  https://github.com/brioche-dev/brioche-packages/pull/843#issuecomment-3092233512.


kylewlacy (approved):
Oh this is awesome!! Really cool seeing how to use nushell for better error handling goo

PR #922, opened 2025-07-28
Add GitHub tags live update

jaudiger:
As discussed earlier, here is the new `std.liveUpdateFromGithubTags()` helper for live update check through GitHub tags. The interface is pretty similar to `std.liveUpdateFromGithubReleases()`. I tweaked the previous Nushell script to make it easier to filter per tag:

```nu
# Previous
http get https://api.github.com/repos/asciinema/asciinema/git/matching-refs/tags
      | get ref
      | each {|ref|
        $ref
        | parse --regex '^refs/tags/v(?P<tag>(?P<major>[\\d]+)\\.(?P<minor>[\\d]+)\\.(?P<patch>[\\d]+))$'
        | get -o 0
      }
      | sort-by -n major minor patch

# New
http get https://api.github.com/repos/asciinema/asciinema/git/matching-refs/tags
  | get ref
  | each {|ref|
    $ref
      # 'refs/tags/' is the prefix for tags in GitHub API responses
      # and is followed by the tag name, remove it
      | str substring 10..-1
      | parse --regex $env.matchTag
      | get -o 0
  }
  | sort-by --natural version
```

One package wasn't updated, that's `nodejs` since we handle multiple versions in this one. But I'll update the logic to make it compatible with the new helper in an upcoming PR. It'll then let us enable checking for multiple versions in other packages:

- `go_mockery`
- `python`

### DIFF
--- a/packages/asciinema/project.bri
+++ b/packages/asciinema/project.bri
@@ -1,5 +1,4 @@
 import * as std from "std";
-import { nushellRunnable, type NushellRunnable } from "nushell";
 import python from "python";
 
 export const project = {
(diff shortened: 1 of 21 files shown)


kylewlacy (approved):
LGTM :+1:

PR #974, opened 2025-08-07
chore(live-update): validate matchTag parameter to ensure it contains 'version'

jaudiger:
This PR is a pre-requisite to update std GitHub release live update methods to fetch all the releases instead of the latest one.

`matchTag` is being used in three std live update methods:

- `std.liveUpdateFromGithubReleases()`
- `std.liveUpdateFromGithubTags()`
- `std.liveUpdateFromGitlabReleases()`

In the methods based on release, the parameter `matchTag` was validated in the Nushell script, but not in the method based on tag. This PR extracts this validation to move it in Typescript. The advantage of doing it this way is, we can now emit at runtime a proper reminder to the end user that if he overrides the default regex used to match tags, this regex must contains a `version` field.

### DIFF
--- a/packages/std/extra/live_update/from_github_releases.bri
+++ b/packages/std/extra/live_update/from_github_releases.bri
@@ -1,6 +1,9 @@
 import * as std from "/core";
 import { parseGithubRepo } from "./github_global.bri";
-import { DEFAULT_LIVE_UPDATE_REGEX_VERSION_MATCH } from "./index.bri";
+import {
+  DEFAULT_LIVE_UPDATE_REGEX_VERSION_MATCH,
+  validateMatchTag,
+} from "./index.bri";
 import type {} from "nushell";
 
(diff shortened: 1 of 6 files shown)


jaudiger:
Here is an example of the new error:

```bash
Error: error when calling liveUpdate

Caused by:
    Error: matchTag must declare a named "version" group
        at validateMatchTag (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/std/extra/live_update/index.bri:23:11)
        at Module.liveUpdateFromGithubReleases (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/std/extra/live_update/from_github_releases.bri:81:3)
        at Module.liveUpdate (file:///Users/jaudiger/Development/git-repositories/jaudiger/brioche-packages/packages/biome/project.bri:43:14)

⏵ Task `Run package live-update` finished with non-zero error code: 1
```

PR #1013, opened 2025-08-09
Update live update GitHub releases

jaudiger:
`std.liveUpdateFromGithubReleases()` does now fetch all the releases as its name suggests: 'releaseS' and not 'latestRelease'. As of today, this function was only retrieving the latest GitHub Release for a package but it caused issues in the past since for a few packages. It can happen that the latest release is not a final release (see https://github.com/asciinema/asciinema/releases/tag/v3.0.0-rc.5 for example). This PR does not necessarily resolved this specific issue, since `std.liveUpdateFromGithubReleases()` and `std.liveUpdateFromGithubTags()` (without `matchTag` overriding) will both return the wrong tag -> the real issue comes from the maintainers not having tagged properly the tag v3.0.0-rc.5 as a pre-release. But now, we can move from `std.liveUpdateFromGithubTags()` to `std.liveUpdateFromGithubReleases()` when releases are available for a package, behind the scene, both methods have now a similar behavior -> both of them now retrieved a list of tags and not the latest one.

Bonus point: we can now include pre-released versions when fetching the list of releases with `std.liveUpdateFromGithubReleases()`

### DIFF
--- a/packages/asciinema/project.bri
+++ b/packages/asciinema/project.bri
@@ -102,9 +102,7 @@ export async function test(): Promise<std.Recipe<std.File>> {
 }
 
 export function liveUpdate(): std.Recipe<std.Directory> {
-  // TODO: to be later replaced with `std.liveUpdateFromGithubReleases()`, once
-  // GitHub releases live method is able to retrieve a list of releases
-  return std.liveUpdateFromGithubTags({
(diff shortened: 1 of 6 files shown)


kylewlacy (approved):
LGTM

PR #1015, opened 2025-08-10
Refactor GitHub Actions workflows using merge queues

### LINKED ISSUE #962: Set up GitHub merge queue
We frequently have builds that fail while auto-merging live-update PRs. Occasionally, two PRs will succeed individually, but will fail when merged together.

GitHub PR merge queues are designed to address problems like this. We should update our CI pipeline to use a merge queue.

### LINKED ISSUE #963: Fix failing live-update PRs sometimes auto-merging
We've had a few cases where a live-update PR build failed, but the PR was auto-merged anyway. A few examples:

- https://github.com/brioche-dev/brioche-packages/pull/892
- https://github.com/brioche-dev/brioche-packages/pull/944

kylewlacy:
Closes #962, closes #963

**Need to do before merging**
- [ ] Enable merge queues in repo settings
- [ ] Update required status checks in repo settings

---

This PR updates our GitHub Actions workflows to use GitHub's [merge queues](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue) (see #962).

With this change, PRs no longer get merged directly to `main`. Instead, they get queued to merge. Then, GitHub Actions will trigger the `merge_group` event for each PR in the queue in sequence. Any PRs that have failing status checks will get de-queued and sent back for review/fixes. Successful PRs then get merged (triggering the `push` event, like normal).

Also, there are repo-level settings for "merge groups". You can set a minimum and maximum number of PRs in a group (between 1 and 100). My understanding is that the merge queue only does the merge after finishing each group (for example, if you merge 100 PRs at the same time and they are all part of the same merge group, GitHub will only trigger the `push` event for the target branch one time after triggering the `merge_group` event for the whole group).

Oh, and finally, I tweaked the "All builds passed" GitHub Actions step to avoid failed builds getting treated as successful (#963). Basically, GitHub will resolve a step as `skipped` if any of its dependencies fail-- which is what happened to the "All builds passed" step, which we were using for our required status checks. To avoid this, the "All builds passed" explicitly fails if any of its dependencies fail using the special `if: always()` condition along with some custom logic.

---

Our workflows now work like this:

- When creating a PR: run `brioche check`
- When a PR enters the merge queue: run `brioche check` and `brioche build`
- On push to `main`: run `brioche check`, `brioche build`, and `brioche publish`

This applies for normal PRs and live-updates. This has the side effect that live-updates no longer auto-build when the PR is opened, but this also leads to less "wasted" builds _and_ helps catch version conflicts earlier (for example, if live-updates are opened for both `ripgrep` and `rust` then `rust` gets merged first, the automatic `ripgrep` build would be wasted because it'll need to rebuild again anyway)

Prior to making the changes in this repo, I did a fair amount of experimenting the the repo https://github.com/brioche-dev/gh-actions-test (N.B.: this repo will probably go away soon).


.github/workflows/_build.yml
```diff
@@ -132,7 +132,10 @@ jobs:
...
   all-builds-passed:
     name: All builds passed
+    if: always() && inputs.enabled && (needs.build.result == 'success' || needs.build.result != 'failed')
     needs: [build]
     runs-on: ubuntu-24.04
     steps:
-      - run: ":"
+      - run: test "$BUILD_RESULT" = "success"
+        env:
+          BUILD_RESULT: ${{ needs.build.result }}
```
jaudiger, line 141: Is there any particular reason to use a env var here ?

jaudiger (approved):
Looks good to me ! I can’t wait to see this modification for the next live update scheduled ! Current opened MRs won’t take advantage of this unless we merge main in them I guess.

.github/workflows/_build.yml
kylewlacy: ...one thing I feel pretty strongly about is that string interpolation should _never_ be used for a script directly in a GitHub Actions workflow[^1]. Instead, I always prefer to inject any dynamic values as env vars and use those, which resolves any problems around string quoting, string expansion, etc.

For a concrete example where this mattered in practice, see [this story about a malicious branch name being used to inject a cryptominer in a project](https://lobste.rs/s/btagmw/maliciously_crafted_github_branch_name). In this particular case, it doesn't take much work to "prove" that this is safe because `needs.<id>.result` never contains special characters, so it's more about good practice (for comparison, I'd still use a SQL bind parameter rather than string interpolation for a SQL query as a default (even if I knew the string was always valid), unless I really _needed_ string interpolation for some reason)

[^1]: The only exception would be is if the interpolated string is itself literally a shell script, or a part of a shell script (I've done this before with matrices, where one of the matrix inputs was a small shell snippet used as a build step)

PR #1088, opened 2025-08-24
Add `coturn` package

kylewlacy:
This PR adds a package for [coturn](https://github.com/coturn/coturn), a free open source implementation of TURN and STUN Server.

This includes basic support, with just `libevent` and `openssl` as dependencies. coturn also optionally supports Prometheus, SQLite, Postgres, and MySQL which aren't enabled in this PR (we have packages for SQLite and Postgres today, but the CMake build didn't pick them up automatically when added as dependencies). Personally, I mainly added this to be able to use the various TURN utils included by default, so I haven't really tried the main server very thoroughly.

### DIFF
--- a/packages/coturn/project.bri
+++ b/packages/coturn/project.bri
@@ -1,0 +1,55 @@
+import * as std from "std";
+import { cmakeBuild } from "cmake";
+import libevent from "libevent";
+import openssl from "openssl";
+
+export const project = {
+  name: "coturn",
+  version: "4.7.0",
+  repository: "https://github.com/coturn/coturn.git",
+};
+
+const source = Brioche.gitCheckout({
+  repository: project.repository,
+  ref: project.version,
+});
+
+export default function coturn(): std.Recipe<std.Directory> {
+  return (
+    cmakeBuild({
+      source,
+      dependencies: [std.toolchain, libevent, openssl],
+    })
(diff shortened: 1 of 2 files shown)


packages/coturn/project.bri
```diff
@@ -0,0 +1,55 @@
...
+});
+
+export default function coturn(): std.Recipe<std.Directory> {
+  return (
+    cmakeBuild({
+      source,
+      dependencies: [std.toolchain, libevent, openssl],
+    })
+      // Fix broken absolute-path symlink
+      .insert("bin/turnadmin", std.symlink({ target: "turnserver" }))
```
kylewlacy, line 24: Not sure why, but the default output includes `turnadmin` as an absolute-path symlink to `turnserver`. This fixes it by using a relative-path symlink instead

packages/coturn/project.bri
```diff
@@ -0,0 +1,55 @@
...
+      .pipe((recipe) =>
+        // Ensure all binaries are marked as executable
+        std.runBash`
+          find bin -type f -exec chmod +x {} +
+        `
```
kylewlacy, line 29: None of the binaries in the output were marked as executable from what I can tell. No clue why that is... so this is a fairly simple workaround

PR #1095, opened 2025-08-25
add package keep-sorted

asheliahut:
This PR adds [keep-sorted](https://github.com/google/keep-sorted) which has a need to use [git] as a dependency during go build to pull the commit information. Also it relies on knowing the current tag in order to create the version command accordingly.

### DIFF
--- a/packages/keep_sorted/project.bri
+++ b/packages/keep_sorted/project.bri
@@ -3,7 +3,7 @@ import git, { gitCheckout } from "git";
 import { goBuild } from "go";
 
 export const project = {
-  name: "keep-sorted",
+  name: "keep_sorted",
   version: "0.7.1",
   repository: "https://github.com/google/keep-sorted.git",
   extra: {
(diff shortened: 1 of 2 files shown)


packages/keep_sorted/project.bri
```diff
@@ -0,0 +1,62 @@
+import * as std from "std";
+import git, { gitCheckout } from "git";
+import { goBuild } from "go";
+
+export const project = {
+  name: "keep-sorted",
```
kylewlacy, line 6: The package name should match the directory name (see: https://github.com/brioche-dev/brioche/issues/67)

```suggestion
  name: "keep_sorted",
```

packages/keep_sorted/project.bri
```diff
@@ -0,0 +1,62 @@
...
+  },
+};
+
+const gitRef = await Brioche.gitRef({
+  repository: project.repository,
+  ref: `v${project.version}`,
+});
+
+const source = gitCheckout(gitRef).pipe((recipe) =>
+  std.runBash`
```
kylewlacy, line 20: #nonblocking Small comment to provide some extra context

```suggestion
  // Create a git tag in the repo, which is used as the version number
  std.runBash`
```

kylewlacy (changes requested):

kylewlacy (approved):
Pushed a few small fixes, LGTM :+1:

PR #1097, opened 2025-08-26
Uniformize python building recipe

jaudiger:
Just to keep things uniformized between recipes. The other recipes were already formatted this way.


kylewlacy (approved):
LGTM, :+1: on figuring out we don't need the `LD_LIBRARY_PATH` setting too!

PR #1099, opened 2025-08-27
fix: resolve a typo in TSDoc comments

jaudiger:
Resolve a typo seen in TSDoc comments of `nushellRunnable` and `bashRunnable`

### DIFF
--- a/packages/nushell/run_nushell.bri
+++ b/packages/nushell/run_nushell.bri
@@ -10,7 +10,7 @@ import nushell from "/";
  * extra dependencies or environment variables using `.dependencies()`
  * or `.env()`, respectively, along with other process options.
  *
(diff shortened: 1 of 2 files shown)


packages/nushell/run_nushell.bri
```diff
@@ -10,7 +10,7 @@ import nushell from "/";
  * extra dependencies or environment variables using `.dependencies()`
  * or `.env()`, respectively, along with other process options.
  *
- * @description See also `nushellRunnable,` which will return a recipe that
+ * @description See also `nushellRunnable` which will return a recipe that
```
jaudiger, line 13: Before merging, should we have here `std.nushellRunnable` as in the other TSDoc comment ?


```suggestion
 * @description See also `std.nushellRunnable` which will return a recipe that
```

PR #1103, opened 2025-08-28
Add the ability to update multiple versions for a recipe

jaudiger:
This PR updates `std.liveUpdateFromGithubTags()` and `std.liveUpdateFromGithubReleases()` to handle multiple versions for a single recipe. It'll be as simple as just updating the `project` variable to include the other possible versions as it's done in the `python` recipe:

```ts
export const project = {
  name: "python",
  version: "3.13.1",
  repository: "https://github.com/python/cpython",
  extra: {
    otherVersions: {
      "3.13": "3.13.1",
      "3.12": "3.12.8",
    },
  },
};
```

Everything will be handled behind the scene when calling `std.liveUpdateFromGithubTags()` and `std.liveUpdateFromGithubReleases()`. The versions will be scraped from the Github API and only the most recent versions will be kept per supported version.


kylewlacy (approved):
Ooh, cool!

PR #1142, opened 2025-08-30
feat(go): update recipe to always support the two latest versions

jaudiger:
It's now fairly easy to add multiple versions support for packages hosted on Github. I took this opportunity to (at least) support the two latest versions of the `go` language. Right now, it doesn't bring any value for the other recipes of Brioche, but if someone wants to build his own recipe (from a private project) locally with the version N-1 of `go`, now he can. And if I'm looking in the future, this could bring values when Brioche will support dev environment, since a user could then decide which version to install for a language.


kylewlacy (approved):
LGTM

I still think that we should add proper runtime support for multiple versions long-term (see https://github.com/brioche-dev/brioche/issues/97), and I think that might look somewhat different than what we're doing in `brioche-packages`. But I'm not sure how soon that feature will happen, so I am happy with this change: language upgrades can definitely break other packages, and we have a lot of packages that use Go!

jaudiger:
@kylewlacy Would like from me to at least adopt a few recommendations from the MVP of [the issue](https://github.com/brioche-dev/brioche/issues/97) ? So something like that:

```ts
export const project = {
  name: "go",
  version: "1.25.0",
  versions: {
    "1.24": "1.24.6",
  },
  repository: "https://github.com/golang/go",
};
```

Indeed there is no need to duplicate the latest version as it currently done:

```ts
{
  version: "1.25.0",
  versions: {
    "1.24": "1.24.6",
  },
}

// Versus

{
  version: "1.25.0",
  versions: {
    "1.25": "1.25.0"
    "1.24": "1.24.6",
  },
}
```

So, both `brioche install -r trunk` and `brioche install -r trunk@latest` could use the `version` field. Whereas `brioche install -r trunk@1.24` would use one of the key of the `versions` field to obtain the real version. I'm not quite sure, it would be interesting to make a differentiation by including the patch version. As in the other package manager, they always tend to use the combo:

- MAJOR
- or, MAJOR.MINOR
-  but not MAJOR.MINOR.MATCH

About the `alpha / beta / nightly / ...`, we could either have it in the key or the value of the key could be extended to be an object:

The advantage to have it directly in the key is it would simplify later the system to choose the correct version, since invoking `brioche install -r trunk@1.26-alpha` would be then straightforward if we only need to look up the keys of `versions`.

jaudiger:
Ok, I'll draft a PR in the coming days, based on to begin:

```ts
export const project = {
  name: "go",
  version: "1.25.0",
  versions: {
    "1.24": "1.24.6",
  },
  repository: "https://github.com/golang/go",
};
```

PR #1149, opened 2025-09-05
refactor(recipes): add missing make commands when building packages

jaudiger:
For reference: https://github.com/brioche-dev/brioche-packages/discussions/230

I tested each recipe one by one by running `brioche build -e test -p packages/RECIPE` to ensure no regression happened. On my machine (MacBook Pro with M1 CPU), `nproc` returns 8. I can definitively see the build improvement here in the updated recipes, the build time is most of the time significantly reduced.


kylewlacy (approved):
🎉

PR #1194, opened 2025-09-06
feat: update make commands to use $(nproc) for parallel builds

jaudiger:
Here is the second part of https://github.com/brioche-dev/brioche-packages/pull/1149, this time targeting the existing `make` commands.

I didn't test all of them, but I did test the most important recipes (cmake, git, linux, etc).

PR #1195, opened 2025-09-07
Improve cmake building

jaudiger:
As mentioned in #1194, this PR is the final work on parallel build in our recipe when using `make` or `cmake` tools. By default for `cmakeBuild()`, parallel build will be enabled, if for a recipe we need to disable it, I updated the TSDoc of this function to state that we only need to set `CMAKE_BUILD_PARALLEL_LEVEL` to `1`.

We can still customise (override) the number of jobs used by CMake by setting this env var as we continue to do in:

- binaryen
- libxc
- llvm

For `pstack`, I removed/unset the env var since the building is fast enough (less than 30s) with 8 threads (on my computer).

### DIFF
--- a/packages/cmake/project.bri
+++ b/packages/cmake/project.bri
@@ -133,11 +133,14 @@ export type CMakeVariableType =
  * Generate and build a CMake project.
  *
  * @description A project build system will be generated out-of-tree from the source project,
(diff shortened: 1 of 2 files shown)


jaudiger:
With this PR, approximatively, 20 more recipes will now use parallel building !

PR #1198, opened 2025-09-09
Rewrite libxml2 recipe

jaudiger:
Rewrite the `libxml2` recipe to use the modern CMake tool building. No difference should be seen when using the library.


kylewlacy (approved):
:+1:

PR #1248, opened 2025-09-19
[rust] Upgrade to v1.90.0

kylewlacy:
[Rust v1.90.0](https://blog.rust-lang.org/2025/09/18/Rust-1.90.0/) was released ~today.

One of the big new features was switching to `ld.lld` by default. Since we use wrappers for linker shenanigans, our Rust package needed some tweaks to wrap the included `ld.lld` executable appropriately.

For now, I just ended up copy/pasting the `useBriocheLd` function from [`std/toolchain/utils.bri`](https://github.com/brioche-dev/brioche-packages/blob/c5956d746351635dd0a2e6596dc3d92ccb8d0a89/packages/std/toolchain/utils.bri#L33). It's a pretty specialized function, so I opted to copy/paste rather than including it in the public interface of the `std` package, at least for the time being (...although I'm not really opposed to just making `useBriocheLd` public from `std`, even if the interface isn't perfect now-- I might do a follow-up).

Locally, I tested with `brioche build -p packages/ripgrep -e test` and `brioche build -p packages/nushell -e test`, both seemed to work without any problems, so it seems like this is in a good place to try and get in.

### DIFF
--- a/packages/rust/brioche.lock
+++ b/packages/rust/brioche.lock
@@ -1,9 +1,9 @@
 {
   "dependencies": {},
   "downloads": {
-    "https://static.rust-lang.org/dist/channel-rust-1.89.0.toml": {
+    "https://static.rust-lang.org/dist/channel-rust-1.90.0.toml": {
       "type": "sha256",
-      "value": "fbd1662e100e7b305908ece23b441cb7534eadfa6336c5f173ff08d1cec174a1"
+      "value": "489c19f20d331765ab2835661eb546de90f6446a107a8db83045e7371e45cae2"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


jaudiger:
Just in time before the live update job of today ! Thanks @kylewlacy, I also tested to build `ripgrep` and `nushell` in my aarch64 environment, and I didn't see any linked issues, which is kind of expected.

Did you know on aarch64 and wasm, Rust toolchain was already making use of `lld` ?

Also for the posterity, here is a comparison between `ld` and `lld` for `ripgrep` package on x86_64 (from https://www.phoronix.com/news/Rust-1.90-LLD-Linking):

image

kylewlacy:
> Did you know on aarch64 and wasm, Rust toolchain was already making use of `lld` ? (for context: https://blog.rust-lang.org/2025/09/01/rust-lld-on-1.90.0-stable/)

Ahh, didn't realize that, interesting! Glad to see the improvements make it to other platforms too (also kinda surprising that we didn't see any issues with aarch64 before... maybe it was falling back to the system linker or something? Not sure)

PR #1286, opened 2025-09-20
Add a new std function to ensure libtool files do not contain absolute deps path

jaudiger:
Add a new std function (inspired from the post process of `acl` recipe of the std toolchain) to ensure libtool files do not contain absolute deps path inside the variable `dependency_libs`. I discovered all of that by working on packaging `libx11` locally (and it now work 🚀), but I first need to upstream a few updates around `libxau` and `libxcb` to make use of the new std method related to libtool files.

Before:

```bash
> cat /tmp/output/lib/libxcb-composite.la
# libxcb-composite.la - a libtool library file
# Generated by libtool (GNU libtool) 2.4.7
#
# Please DO NOT delete this file!
# It is necessary for linking the library.

# The name that we can dlopen(3).
dlname='libxcb-composite.so.0'

# Names of this library.
library_names='libxcb-composite.so.0.0.0 libxcb-composite.so.0 libxcb-composite.so'

# The name of the static archive.
old_library='libxcb-composite.a'

# Linker flags that cannot go in dependency_libs.
inherited_linker_flags=''

# Libraries that this one depends upon.
dependency_libs=' //lib/libxcb.la -L/home/brioche-runner-eaf315e60b9ee608cdcfe57271bb2c755cbe20586c1aebceafe6f992958ff40c/.local/share/brioche/locals/7030701dd55d14702c05c7382ff7ab39d84601640d2c4a06e5d15660083564b7/lib/pkgconfig/../../lib //lib/libXau.la'

# Names of additional weak libraries provided by this library
weak_library_names=''

# Version information for libxcb-composite.
current=0
age=0
revision=0

# Is this an already installed library?
installed=yes

# Should we warn about portability when linking against -modules?
shouldnotlink=no

# Files to dlopen/dlpreopen
dlopen=''
dlpreopen=''

# Directory that this library needs to be installed in:
libdir='//lib'
```

After:

```bash
> cat /tmp/output/lib/libxcb-composite.la
# libxcb-composite.la - a libtool library file
# Generated by libtool (GNU libtool) 2.4.7
#
# Please DO NOT delete this file!
# It is necessary for linking the library.

# The name that we can dlopen(3).
dlname='libxcb-composite.so.0'

# Names of this library.
library_names='libxcb-composite.so.0.0.0 libxcb-composite.so.0 libxcb-composite.so'

# The name of the static archive.
old_library='libxcb-composite.a'

# Linker flags that cannot go in dependency_libs.
inherited_linker_flags=''

[...]

Issue #135, opened 2024-10-19
[std] Add a utility for fixing pkg-config files


kylewlacy:
Lots of packages generate pkg-config files at build time, usually written to `${prefix}/lib/pkgconfig`. These pkg-config files basically always seem to have a variable called `prefix`, which is set to the build-time prefix and is used as a base path for the rest of the variables in the pkg-config file.

Here's an example from the Curl build, under `lib/pkgconfig/curl.pc` (as of commit [`45d2638`](https://github.com/brioche-dev/brioche-packages/commit/45d2638f5edb99b52af8d721913486cdef27877b)):

```pkgconfig
prefix=/
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
supported_protocols="DICT FILE FTP FTPS GOPHER GOPHERS HTTP HTTPS IMAP IMAPS IPFS IPNS MQTT POP3 POP3S RTSP SMB SMBS SMTP SMTPS TELNET TFTP"
supported_features="alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Largefile libz NTLM SSL threadsafe TLS-SRP UnixSockets zstd"

Name: libcurl
URL: https://curl.se/
Description: Library to transfer files with ftp, http, etc.
Version: 8.9.1
Requires:
Requires.private: zlib,libzstd,openssl
Libs: -L${libdir} -lcurl
Libs.private: -lssl -lcrypto -lssl -lcrypto -lzstd -lz
Cflags: -I${includedir}
```

> Note: license omitted, see [`COPYING`](https://github.com/curl/curl/blob/6847733191f3a8d4c475d2a1a8439f6a7edfcd77/COPYING)

You can actually use the `${pcfiledir}` variable to get the directory containing the `.pc` file itself, so it's possible to tweak a pkg-config file to work with portable dirs. For example, we could patch the first line of Curl like this:

```
prefix=${pcfiledir}/../../
```

We already have a few packages that manually patch their pkg-config files with a script:

- [`std.toolchain()`](https://github.com/brioche-dev/brioche-packages/blob/45d2638f5edb99b52af8d721913486cdef27877b/packages/std/toolchain/native/index.bri#L286)
- [`sqlite`](https://github.com/brioche-dev/brioche-packages/blob/45d2638f5edb99b52af8d721913486cdef27877b/packages/sqlite/project.bri#L90)
[...]

PR #1287, opened 2025-09-21
Tweak pkgconfig files

jaudiger:
See https://github.com/brioche-dev/brioche-packages/pull/1286 for context. This PR adds the missing calling to `std.pkgConfigMakePathsRelative` where the recipe contains a pkgconfig folder.

This PR is interesting for `scdoc`, since it's used by `kmod` which is in turn used by `linux` recipe.

### DIFF
--- a/packages/kmod/project.bri
+++ b/packages/kmod/project.bri
@@ -24,7 +24,19 @@ export default function kmod(): std.Recipe<std.Directory> {
   `
     .workDir(source)
     .dependencies(std.toolchain, meson, ninja, scdoc, openssl)
-    .toDirectory();
+    .toDirectory()
+    .pipe(
+      std.pkgConfigMakePathsRelative,
+      (recipe) =>
+        std.setEnv(recipe, {
+          CPATH: { append: [{ path: "include" }] },
+          LIBRARY_PATH: { append: [{ path: "lib" }] },
+          PKG_CONFIG_PATH: {
+            append: [{ path: "lib/pkgconfig" }, { path: "share/pkgconfig" }],
+          },
+        }),
+      (recipe) => std.withRunnableLink(recipe, "bin/kmod"),
+    );
 }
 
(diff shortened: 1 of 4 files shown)


packages/kmod/project.bri
```diff
@@ -24,7 +24,19 @@ export default function kmod(): std.Recipe<std.Directory> {
...
-    .toDirectory();
+    .toDirectory()
+    .pipe(
+      std.pkgConfigMakePathsRelative,
+      (recipe) =>
+        std.setEnv(recipe, {
+          CPATH: { append: [{ path: "include" }] },
+          LIBRARY_PATH: { append: [{ path: "lib" }] },
+          PKG_CONFIG_PATH: {
+            append: [{ path: "lib/pkgconfig" }, { path: "share/pkgconfig" }],
```
jaudiger, line 35: This recipe has two pkgconfig folders which contain different files, so I have added both of them.

packages/postgresql/project.bri
```diff
@@ -37,12 +37,15 @@ export default function postgresql(): std.Recipe<std.Directory> {
...
-      }),
+    .pipe(
+      std.pkgConfigMakePathsRelative,
+      (recipe) =>
+        std.setEnv(recipe, {
+          LIBRARY_PATH: { append: [{ path: "lib" }] },
+          CPATH: { append: [{ path: "include" }] },
+          PKG_CONFIG_PATH: { append: [{ path: "lib/pkgconfig" }] },
+        }),
+      (recipe) => std.withRunnableLink(recipe, "bin/postgres"),
```
jaudiger, line 48: Here I only added the runnable binary

jaudiger:
This kind of resolve this issue: https://github.com/brioche-dev/brioche-packages/issues/135 if this PR is merged, since we'll have ensured that all the recipes are correct.

PR #1288, opened 2025-09-21
Tweak libtool files

jaudiger:
Based on #1286 

See #1286 for context. This PR makes use of the new `std.libtoolSanitizeDependencies` method inside the recipes containing libtool files.

### DIFF
--- a/packages/alsa_lib/project.bri
+++ b/packages/alsa_lib/project.bri
@@ -21,12 +21,15 @@ export default function alsaLib(): std.Recipe<std.Directory> {
     .workDir(source)
     .dependencies(std.toolchain)
     .toDirectory()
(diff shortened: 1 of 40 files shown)


jaudiger:
Waiting first on #1286 and #1287, once both are merged, I'll resolve the merge conflicts. Plus, I'll update this PR if the name of the method changed (see https://github.com/brioche-dev/brioche-packages/pull/1286#issuecomment-3315951457)

jaudiger:
@kylewlacy I didn't add the new std method in the post build step of `cmakeBuild()` (as we do for `std.pkgConfigMakePathsRelative`). As of today, no recipe contains libtool files when building with CMake which is quite expected I guess, since both serves the same purpose (aka finding dependencies for building and linking).

PR #1392, opened 2025-10-03
build(bubblewrap): remove CMake dependency from the build process

jaudiger:
It was unused by the build process of bubblewrap.


kylewlacy (approved):
:+1:

PR #1480, opened 2025-10-16
feat: add pull request templates for adding and updating recipes

jaudiger:
I think only me will use it at the beginning since there is no interface for this "feature" from GitHub. We can define multiple templates for the PR, but these templates can only be accessed through URI query, and cannot be used directly from the interface (which is a bit weird...), see https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository.

So, here is an example of both templates:

- Add a new recipe: https://github.com/jaudiger/test-brioche-repo/compare/main...new-branch?quick_pull=1&template=add_recipe.md
- Update a recipe: https://github.com/jaudiger/test-brioche-repo/compare/main...new-branch?quick_pull=1&template=update_recipe.md

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/add_recipe.md
+++ b/.github/PULL_REQUEST_TEMPLATE/add_recipe.md
@@ -1,0 +1,54 @@
+# Add a new Brioche recipe
+
+## Summary
+
+This Pull Request adds a new recipe to the registry.
+
+- **Recipe name:** `TODO: replace with name`
+- **Website / repository:** `TODO: link`
+- **Short description:** `TODO: one-line description`
+
+# Related issue(s) or discussion(s)
+
+`TODO: Link to the issue or discussion if relevant.`
+
+## Checklist (required)
+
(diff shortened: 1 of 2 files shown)


kylewlacy (approved):
LGTM

Might be fun to eventually have a script or something to submit a PR automatically using the template and filling in e.g. test output!

jaudiger:
> Might be fun to eventually have a script or something to submit a PR automatically using the template and filling in e.g. test output!

Indeed, it could be a good idea to do ! I already have [a script](https://github.com/jaudiger/personal-scripts/blob/main/brioche-wrapper) to validate more efficiently the live update PRs, would you like a script based on Nushell or Bash or something else ? Or should we wait the support of `cargo-script` (example [here](https://github.com/rust-lang/cargo/issues/12207#issuecomment-3412997290)), but I don't know when it will land on Rust stable, though 😄

PR #1532, opened 2025-10-23
Enhance live update capabilities

jaudiger:
This PR is the continuation of the work to add new sources for live updates checking.

It adds four new sources and puts Gitea and GitLab at (almost) the same quality level than GitHub integration, we're still missing a few additional capabilities such as the support of `extra.otherVersions`, `extra.versionUnderscore` and `extra.versionDash`, but it could be added in the future if we need the support of these variables:

New sources:

- Gitea tag
- GitLab tag
- Go module
- Python package

As a follow up, I could update the function `liveUpdate()` to use the right source depending on each recipe.

All of this work is part of using the right tool for the right use case, and this work will serve as a foundation for the documentation.

PS: I have tested each new source locally, but I didn't include any integration in this PR.


kylewlacy (approved):
LGTM 💯

PR #1535, opened 2025-10-24
fix(std): resolve typo in autopack options documentation

jaudiger:
Resolved another typo I saw recently. Sorry I saw them one by one


kylewlacy (approved):
Always good to get typos resolved! Luckily our "no op" builds (like when we have formatting or comment changes) are a lot faster to get through CI than they used to be

PR #1536, opened 2025-10-24
fix(std): ensure post build methods do follow symbolic links

jaudiger:
This is another rebuild the world PR, but I saw the issue while working on #1533.

The `json_c` package pushes all its output into `lib64` which is then symlinked to `lib` at the end of `cmakeBuild()`. Unfortunately, our post build methods do not follow the links of such files. So pkg config and libtool files weren't updated.

Also, this PR fixes a possible issue with `LIBTOOL_SANITIZE_DEPENDENCIES_SCRIPT`, previously, if the recipe wasn't containing any `*.la` files, an error was thrown. This is now handled more gracefully through the usage of `find first -> then execute command`.

Before:

```
1786204│ sed: can't read /home/brioche-runner-ea6ff3839e11210c3e3f1d16dda0739451a7cf14a5741af642e2f4038e57776f/.local/share/brioche/outputs/output-ea6ff3839e11210c3e3f1d16dda0739451a7cf14a5741af642e2f4038e577
       │ 76f/lib/*.la: No such file or directory
```

After

```
... no issue
```

### DIFF
--- a/packages/std/extra/libtool_sanitize_dependencies.bri
+++ b/packages/std/extra/libtool_sanitize_dependencies.bri
@@ -2,7 +2,16 @@ import * as std from "/core";
 import { runBash } from "./run_bash.bri";
 
 export const LIBTOOL_SANITIZE_DEPENDENCIES_SCRIPT = std.indoc`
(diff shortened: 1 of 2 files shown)


kylewlacy (approved):
:+1:

PR #1589, opened 2025-10-27
Use more and more pipes in recipes

jaudiger:
The work on the PR has been separated in "unit work" for better reviewing on what was done.

Here is a quick summary:

- Post build work of `cmakeBuild()` is now separated into a post build step, since it can be considered as outside of the build process
- cd commands in Nushell scripts (for `nodejs` and `pnpm`) have been suppressed in favour of `std.process` worker option
- `std.pipe` is used in more places


kylewlacy (approved):
Alright, merging as-is so we can kick off the build. The other PR is up so I'm not sure if merging this actually makes sense now, so sorry if this causes extra churn!

kylewlacy (commented):
~~I did catch one tiny formatting thing~~ **Edit**: Left this comment on the wrong PR, whoops

kylewlacy:
Umm, I removed this PR from the queue because it wouldn't pick up the new commit otherwise, but now the status checks don't seem to be triggering...

PR #1591, opened 2025-10-28
fix(std): update regex to allow exclamation mark in go module name for live update

jaudiger:
[Per specification](https://github.com/golang/vgo/blob/f621561646e8078690ff9966602bf506a52f9475/vendor/cmd/go/internal/modfetch/proxy.go#L56-L59). I saw it while further experimenting with the newly added live update methods.


kylewlacy (approved):
Ooh interesting!

PR #1592, opened 2025-10-28
Experiment with go module source download

jaudiger:
This PR is built on top of #1591, and is here to demonstrate how to fetch Go module sources from the Go registry instead of fetching them from a forge repository. For Go modules, everything could be fetched from the registry:

- the source (of course)
- the build date
- the git commit hash

Why this PR ? Since we have now reached more than 300 packages in the Brioche registry, I would like to take a step back to work on two main tasks by the end of November / December while still working on upstreaming new packages:

- Stability: fix some issues (meson building, ncurses library not found sometimes, nodejs install issue with some platforms, etc) -> things I saw locally by experimenting with some recipes
- Standardization: work on a common set of steps per language / tool that would make sense for someone used to it, and fallback to a generic common set of steps when it's not possible. The idea behind this PR is: for someone used to Go, it could make more sense to directly use the Go proxy to fetch sources, etc (by staying in a known dev loop environment) than by retrieving the source from a forge repository. And if the package is not present in the Go registry, we could still fallback to what we have done so far in Brioche recipes (aka cloning the sources from a forge repository). At the end, I would like to apply this concept for Rust crates, NPM packages, Python packages, etc -> use the language / tool registry if available or fallback to a git repo. Of course, I'll have to properly document the creation of a new Brioche recipe in the newly added [packaging guide](https://brioche.dev/docs/contributing/packaging-guide/) on brioche.dev.

My personal goal for the end of 2025 would be to have a safe and clean foundation for Brioche recipes before trying to reach 500 recipes 😁 in 2026.

### DIFF
--- a/packages/aks_mcp_server/project.bri
+++ b/packages/aks_mcp_server/project.bri
@@ -1,6 +1,5 @@
 import * as std from "std";
-import * as typer from "typer";
-import { ModuleInfo, goBuild } from "go";
+import { goBuild, goModuleManifest } from "go";
 
 export const project = {
   name: "aks_mcp_server",
@@ -15,11 +14,10 @@ const source = Brioche.download(
 )
   .unarchive("zip")
   .peel(3);
-const manifest = typer.parseJson(
-  ModuleInfo,
-  await Brioche.download(
+const manifest = await goModuleManifest(
+  Brioche.download(
     `https://proxy.golang.org/${project.extra.moduleName}/@v/v${project.version}.info`,
-  ).read(),
+  ),
 );
 
 export default function aksMcpServer(): std.Recipe<std.Directory> {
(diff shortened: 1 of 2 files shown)


kylewlacy:
This looks really cool, really neat to see the `.info` file especially! :tada: And definitely on board with getting more stuff in place to try and make packaging new stuff easier!

---

The Go proxy is also a pretty interesting thing too... I forget the specifics but doesn't the Go proxy basically just mirror a GitHub download, or whatever the upstream is? (We're still proxying the download either way through the Brioche cache)

...if so, does that mean that any Go module could be replaced using the upstream download instead and always have it be equivalent?

If that's right, then `crates.io` is kind of the opposite: there's no guarantee that the upstream repo matches the download from `crates.io`, and there's not even a requirement that the upstream project has an open source repo at all, or that the source code matches what's in `crates.io`, so it does make sense to think of `crates.io` as its own data source

Just trying to figure out what how the Go proxy fits in! This has also got me thinking again about if there are things we can do to simplify Go Proxy / crates.io / other data sources, where it'd be better to not force users to deal with these sort-of-internal URLs themselves

jaudiger:
> Just trying to figure out what how the Go proxy fits in! This has also got me thinking again about if there are things we can do to simplify Go Proxy / crates.io / other data sources, where it'd be better to not force users to deal with these sort-of-internal URLs themselves

It would be nice that at some points (as we discussed earlier a few months ago) that we could run some basic typescript methods to extract the URLs to put in the lock file. This way, we could hide all the URLs from the end user into proper methods that are located in revenant locations. As opposed to today where we just basically extract these URLs from the `Brioche.download()`.

kylewlacy (approved):
LGTM, awesome to see first-class support for more and more ecosystems!!

kylewlacy:
Oh, and I also opened an issue about the lockfile / `Brioche.download` thing: https://github.com/brioche-dev/brioche/issues/353

jaudiger:
> (actually I have no idea what the 1761034664 ID is...)

I checked their build process, and I didn't find any mention to `1761034664`.

> @jaudiger any thoughts on handling this? I think it'd be a little better if we could get rid of the leading v, which I guess would mean either stripping it from the manifest.Version field or just using project.version here instead. I'd also be okay keeping it as-is (although that could lead to problems in the future... probably not for this package but this looks like a kind of standard output format, and there could be shell scripts that try to pull the version number from the output of --version that might not handle a leading v)

Right, I'll encapsulate this logic in a function exported from the go module, and the function will strip `v` if it prefixes the build version.

jaudiger:
@kylewlacy I pushed 5f86bfbd28f8f4549a5a1523cc4bc3ab78e9db53 to encapsulate the retrieving of Go module manifest. I could later move this logic into a new file in the `go` recipe to properly split the logic between the `go` recipe and the features around this recipe.

PR #1602, opened 2025-10-31
Do not handle prerelease in live update

jaudiger:
This PR updates the default behaviour for live updates methods by not parsing prerelease versions (as it was discussed [here](https://github.com/brioche-dev/brioche-packages/pull/1566#issuecomment-3460401841)).


kylewlacy (approved):
Awesome!

PR #1603, opened 2025-10-31
Update rust methods

jaudiger:
This PR does two things:

- resolves an issue in `createLockFile` when `currentDir` was set
- use `std.process` where it's applicable

### DIFF
--- a/packages/rust/project.bri
+++ b/packages/rust/project.bri
@@ -369,6 +369,21 @@ export function cargoBuild(
 }
 
 /**
+ * Options for creating a lock file from a Rust crate.
+ *
+ * @param source - The crate to create the lockfile for.
+ * @param currentDir - Optionally set a subpath to use as the starting directory
+ *   relative to `source`. This is useful when building a Cargo workspace within
+ *   a monorepo project.
+ * @param allowLockfileGeneration - Whether to allow generating the lockfile.
+ */
+interface CreateLockFileCrateOptions {
+  source: std.RecipeLike<std.Directory>;
+  currentDir?: string;
+  allowLockfileGeneration: boolean;
+}
+
+/**
  * Generate the Cargo.lock file if it doesn't exist.
  * Warning: if this file is missing from the upstream source, it breaks
  * reproducible builds, but we need it to vendor the dependencies.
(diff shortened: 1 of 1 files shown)


packages/rust/project.bri
```diff
@@ -379,31 +394,36 @@ export function cargoBuild(
  * @return The source crate with the lockfile.
  */
 function createLockfile(
-  crate: std.RecipeLike<std.Directory>,
-  allowLockfileGeneration: boolean,
+  options: CreateLockFileCrateOptions,
 ): std.Recipe<std.Directory> {
-  return std.runBash`
-    if [ "$allowLockfileGeneration" = "true" ]; then
```
jaudiger: The if statement has been extracted, and it is now run in typescript

packages/rust/project.bri
```diff
@@ -425,26 +445,35 @@ interface CreateSkeletonCrateOptions {
...
-  `
-    .dependencies(rust, cargoChef)
-    .env({ recipe, current_dir: currentDir ?? "" })
+
+  return std
+    .process({
+      command: "cargo",
+      args: ["chef", "cook", "--recipe-path", recipe, "--no-build"],
+      dependencies: [rust, cargoChef],
+      outputScaffold:
```
jaudiger: Basically replace the mkdir of the previous `std.bash`

packages/rust/project.bri
```diff
@@ -379,31 +394,36 @@ export function cargoBuild(
...
-        exit 1
-      fi
+  if (!allowLockfileGeneration) {
+    return std.recipe(source);
+  }
 
-      cargo generate-lockfile
+  return std.runBash`
+    # Check if the Cargo.lock file exists, if yes, exit on error
+    if [ -f Cargo.lock ]; then
```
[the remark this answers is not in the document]
kylewlacy: I think using a Bash script here makes the most sense for now. It _might_ be possible to do in TypeScript, but would require `await`ing part of the build, which... well, I'd rather avoid if possible!

I have had an idea kicking around for a while about "condition recipes", where we could represent conditions like this directly within a recipe. I started a discussion to get some of those ideas down here: https://github.com/brioche-dev/brioche/discussions/359

kylewlacy (approved):
LGTM!

PR #1605, opened 2025-11-01
refactor: simplify currentDir logic in Go and Rust project files

jaudiger:
Simplify `currentDir` in `go` and `rust` recipes, by always using "std.tpl\`${std.workDir}/${options.currentDir}\`" instead of a ternary operator. At the end, when `options.currentDir` is null, it'll be the same path whereas we were using `std.workDir` or `std.tpl`${std.workDir}/`. The only downside of this modification is: it'll trigger a rebuild of the recipes that do not set `options.currentDir`.

### DIFF
--- a/packages/go/project.bri
+++ b/packages/go/project.bri
@@ -262,11 +262,7 @@ export function goBuild(options: GoBuildOptions): std.Recipe<std.Directory> {
     go install "\${goargs[@]}" "$package_path"
   `
     .workDir(options.source)
-    .currentDir(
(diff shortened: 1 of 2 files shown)


kylewlacy (approved):
LGTM

It does seem a little silly that this causes rebuilds! I wonder if we should update `std.process` to normalize the `currentDir` template by trying to strip a trailing `/` if present... (I think that's harder than it sounds due to how templates work though!)

PR #1664, opened 2025-11-03
Update live update match tag

jaudiger:
Supersedes https://github.com/brioche-dev/brioche-packages/pull/1645, https://github.com/brioche-dev/brioche-packages/pull/1651, https://github.com/brioche-dev/brioche-packages/pull/1638 and https://github.com/brioche-dev/brioche-packages/pull/1627.

After the merge of https://github.com/brioche-dev/brioche-packages/pull/1602, I changed the default live update for GitHub release to align it with other checking methods:

image

The GitHub release now relies only on the default regex as in the other methods. But with the last live update checking, it broke a few live update recipes (where we were overriding the `matchTag`). The problem mostly comes from our overrides since they were too opened. This PR is here to enforce better constraints when parsing the versions of a package. It's a compromise between a regex that does the job without being to verbose.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/add_recipe.md
+++ b/.github/PULL_REQUEST_TEMPLATE/add_recipe.md
@@ -24,7 +24,7 @@ This Pull Request adds a new recipe to the registry.
 1. Run the **test** scenario:
 
 ```bash
-   brioche build -e test -p RECIPE_PATH
+brioche build -e test -p RECIPE_PATH
 ```
 
 <details><summary>Test output (click to expand)</summary>
--- a/.github/PULL_REQUEST_TEMPLATE/update_recipe.md
+++ b/.github/PULL_REQUEST_TEMPLATE/update_recipe.md
@@ -15,7 +15,7 @@
 1. Run the **test** scenario:
 
 ```bash
-   brioche build -e test -p RECIPE_PATH
+brioche build -e test -p RECIPE_PATH
 ```
 
 <details><summary>Test output (click to expand)</summary>


packages/cargo_nextest/project.bri
```diff
@@ -32,18 +31,16 @@ export async function test(): Promise<std.Recipe<std.File>> {
...
+  std.assert(
+    result.startsWith(expected),
+    `expected '${expected}', got '${result}'`,
+  );
 
   return script;
 }
 
 export async function liveUpdate(): Promise<std.Recipe<std.Directory>> {
-  return std.liveUpdateFromGithubReleases({
```
jaudiger: This is the advantage of checking from the crates.io, we don't need the `matchTag` here.

packages/cargo_nextest/project.bri
```diff
@@ -4,22 +4,21 @@ import rust, { cargoBuild } from "rust";
...
+)
+  .unarchive("tar", "gzip")
+  .peel();
 
 export default function cargoNextest(): std.Recipe<std.Directory> {
   return cargoBuild({
     source,
-    path: "cargo-nextest",
-
-    // cargo-chef makes the lockfile out-of-date, see this issue:
```
jaudiger: It seems we don't need anymore this hack here.

packages/cargo_nextest/project.bri
```diff
@@ -4,22 +4,21 @@ import rust, { cargoBuild } from "rust";
...
-    // https://github.com/LukeMathWalker/cargo-chef/issues/305
-    cargoChefPrepare: false,
+    runnable: "bin/cargo-nextest",
```
jaudiger: I also added `runnable`. It has nothing to do with the other changes of this PR.

kylewlacy (approved):
LGTM

jaudiger:
Ok, I resolved the merge conflicts, since you approved this PR, I'll put in the merge queue directly.

Issue #838, opened 2025-07-15
[meson] Fix invalid Python path being used


kylewlacy:
Originally discussed on Zulip: https://brioche.zulipchat.com/#narrow/channel/440653-general/topic/ninja.20dependency.20unable.20to.20find.20python.20dependency.3F/near/528639997

Based on a build error while packaging [pixman](), `jokeyrhyme` got a build error. Here was the `project.bri` file:

```typescript
import * as std from "std";
import meson from "meson";
import ninja from "ninja";
import python from "python";

export const project = {
  name: "pixman",
  version: "0.46.2",
  repository: "https://gitlab.freedesktop.org/pixman/pixman.git",
};

const source = Brioche.gitCheckout({
  repository: project.repository,
  ref: `${project.name}-${project.version}`,
});

export default function pixman(): std.Recipe<std.Directory> {
  return std.runBash`
      python --version
      meson setup build --prefix /
      meson compile -C build
      meson install -C build --destdir "$BRIOCHE_OUTPUT"
    `
    .dependencies(std.toolchain, meson, ninja, python)
    .workDir(source)
    .toDirectory();
}
```

...and a snippet of the build error:

```
FAILED: [code=127] pixman/libpixman-1.so.0.46.2.p/libpixman-1.so.0.46.2.symbols
/home/brioche-runner-6d4911c2dbdb66ac913b001710267c670ddfe3137131966368cb580e9334b4b6/.local/share/brioche/locals/58d22139dc97313596a0fa3cd130b357bd8996fae68c6c3534ce1aeb1f36f97f/venv/bin/meson --internal symbolextractor /home/brioche-runner-6d4911c2dbdb66ac913b001710267c670ddfe3137131966368cb580e9334b4b6/work/build pixman/libpixman-1.so.0.46.2 pixman/libpixman-1.so.0.46.2 pixman/libpixman-1.so.0.46.2.p/libpixman-1.so.0.46.2.symbols/home/brioche-runner-6d4911c2dbdb66ac913b001710267c670ddfe3137131966368cb580e9334b4b6/.local/share/brioche/locals/58d22139dc97313596a0fa3cd130b357bd8996fae68c6c3534ce1aeb1f36f97f
[...]

PR #1669, opened 2025-11-04
Update shebang in Python venv scripts

jaudiger:
Resolves https://github.com/brioche-dev/brioche-packages/issues/838.

This PR fixes the shell scripts generated in the Python venv environment when installing a Python package. The shebang in the shell scripts were referencing a **fixed** absolute path to the Python interpreter.

Here there are two things done:

- Switch to a **dynamic** path for the Python interpreter (by using `"$(dirname "$0")/python"`)
- The scripts are updated to be [polyglot](https://en.wikipedia.org/wiki/Polyglot_(computing)) to ensure they can run both as a shell script and also a python script, without any modification needed

Relying on polyglot script, let's us simplify the adding of runnable to the recipe since we don't need anymore to point to the Python interpreter from the venv environment:

To take an example, here is the before:

```ts
std.addRunnable(recipe, "bin/meson", {
  command: { relativePath: "venv/bin/python" },
  args: [{ relativePath: "venv/bin/meson" }],
})
```

And the after with a polyglot script:

```ts
std.addRunnable(recipe, "bin/meson", {
  command: { relativePath: "venv/bin/meson" },
})
```

Finally, here is the before and after for the meson shell script:

```
> cat /tmp/output/venv/bin/meson
#!/bin/sh
'''exec' /home/brioche-runner-e0b40aa08a68f897290b857707c3175a3c4c7599b07664d125b58d2289bae146/.local/share/brioche/outputs/output-e0b40aa08a68f897290b857707c3175a3c4c7599b07664d125b58d2289bae146/bin/python "$0" "$@"
' '''
import sys
from mesonbuild.mesonmain import main
if __name__ == '__main__':
    if sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][:-4]
    sys.exit(main())

> cat /tmp/output/venv/bin/meson
#!/bin/sh
'''exec' "$(dirname "$0")/python" "$0" "$@"
' '''
import sys
from mesonbuild.mesonmain import main
if __name__ == '__main__':
    if sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][:-4]
    sys.exit(main())
```

### DIFF
--- a/packages/aws_cli/project.bri
+++ b/packages/aws_cli/project.bri
@@ -1,5 +1,5 @@
 import * as std from "std";
-import python, { fixScriptShebangs } from "python";
+import python, { fixScriptShebangs, wrapVenvScripts } from "python";
 
 export const project = {
   name: "aws_cli",
@@ -96,10 +96,9 @@ export default function awsCli(): std.Recipe<std.Directory> {
       venv,
     })
     .pipe((recipe) =>
-      // Add `bin/aws` to run the main script
-      std.addRunnable(recipe, "bin/aws", {
-        command: { relativePath: "venv/bin/python" },
-        args: [{ relativePath: "venv/bin/aws" }],
+      wrapVenvScripts(recipe, {
+        venvDir: "venv",
+        scripts: ["aws", "aws_completer"],
       }),
     )
     .pipe((recipe) => std.withRunnableLink(recipe, "bin/aws"));
--- a/packages/meson/project.bri
+++ b/packages/meson/project.bri
@@ -1,5 +1,5 @@
 import * as std from "std";
-import python from "python";
+import python, { wrapVenvScripts } from "python";
 
 export const project = {
   name: "meson",
@@ -53,20 +53,14 @@ export default function meson(): std.Recipe<std.Directory> {
     .toDirectory();
 
   // Create the final recipe with the venv under `venv`
-  return (
-    std
-      .directory({
-        venv,
-      })
-      // Add `bin/meson` to run the main script
-      .pipe((recipe) =>
-        std.addRunnable(recipe, "bin/meson", {
-          command: { relativePath: "venv/bin/python" },
-          args: [{ relativePath: "venv/bin/meson" }],
-        }),
-      )
-      .pipe((recipe) => std.withRunnableLink(recipe, "bin/meson"))
-  );
+  return std
+    .directory({
+      venv,
+    })
+    .pipe((recipe) =>
+      wrapVenvScripts(recipe, { venvDir: "venv", scripts: ["meson"] }),
+    )
+    .pipe((recipe) => std.withRunnableLink(recipe, "bin/meson"));
 }
 
 export async function test(): Promise<std.Recipe<std.File>> {


kylewlacy (approved):
LGTM :tada: 

I feel like this is the best option we have for now for wrapping Python scripts. My feeling is there's no _perfect_ solution sadly (this approach requires having `/bin/sh` available, may not be true e.g. in a Docker container-- using a binary-based wrapper could no longer be called with `python` directly), but I like this as the direction to go for Python. Really excited to finally close #838!!

jokeyrhyme:
wow!

> this approach requires having /bin/sh available, may not be true

could a future PR have Python explicitly depend upon and be provided with a path to `sh` ?

kylewlacy:
> could a future PR have Python explicitly depend upon and be provided with a path to `sh` ?

@jokeyrhyme I think using `#!/usr/bin/env sh` might be slightly better, and would mean we could use whatever `sh` is on the `$PATH`... but that would still depend on `/usr/bin/env` of course. We _could_ add support to just pass an arbitrary path (e.g. making it an argument so you could pass `/usr/local/bin/bash` if you wanted), but I think that would complicate the code a fair amount (so I'd want to discuss real-world use-cases before going that direction, I think)

The other direction we could go is to use a packed executable, similar to what we do with Bash scripts and other scripts. Basically, we'd wrap the Python scripts with a small binary that, in turn, calls the right Python interpreter. The advantage is it doesn't depend on any absolute paths, but the disadvantage is that the "script" is no longer a plaintext Python script, meaning you couldn't just call it with e.g. `python ./bin/meson`. You'd _have_ to call it by doing `./bin/meson` directly instead.

jaudiger:
Just for the comparison between the before and after this PR:

- AWS (before)

<details>

```bash
#!/home/brioche-runner-99ff7916c9d87b5596181b71d7f0a2ef6650778f0a1e4f8d6cb0ee29b21c8ca6/.local/share/brioche/outputs/output-99ff7916c9d87b5596181b71d7f0a2ef6650778f0a1e4f8d6cb0ee29b21c8ca6/bin/python
# Copyright 2012 Amazon.com, Inc. or its affiliates. All Rights Reserved.

# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at

#     http://aws.amazon.com/apache2.0/

# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
import os
import sys

if os.environ.get('LC_CTYPE', '') == 'UTF-8':
    os.environ['LC_CTYPE'] = 'en_US.UTF-8'
import awscli.clidriver


def main():
    return awscli.clidriver.main()


if __name__ == '__main__':
    sys.exit(main())
```

</details>

- Meson (before)

<details>

```bash
#!/bin/sh
'''exec' /home/brioche-runner-bce7cee8aad8d382153d96df628e58cd8d5a09c7b81979b688c35e303ee14c35/.local/share/brioche/outputs/output-bce7cee8aad8d382153d96df628e58cd8d5a09c7b81979b688c35e303ee14c35/bin/python "$0" "$@"
' '''
import sys
from mesonbuild.mesonmain import main
if __name__ == '__main__':
    if sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][:-4]
    sys.exit(main())
```

</details>

- AWS (after)

<details>

```bash
#!/bin/sh
'''exec' "$(dirname "$0")/python" "$0" "$@"
' '''
# Copyright 2012 Amazon.com, Inc. or its affiliates. All Rights Reserved.

# Licensed under the Apache License, Version 2.0 (the "License"). You
# may not use this file except in compliance with the License. A copy of
# the License is located at

#     http://aws.amazon.com/apache2.0/

# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
# ANY KIND, either express or implied. See the License for the specific
# language governing permissions and limitations under the License.
import os
import sys

if os.environ.get('LC_CTYPE', '') == 'UTF-8':
    os.environ['LC_CTYPE'] = 'en_US.UTF-8'
import awscli.clidriver


def main():
    return awscli.clidriver.main()


if __name__ == '__main__':
    sys.exit(main())
```

</details>

- Meson (after)

<details>

```bash
#!/bin/sh
'''exec' "$(dirname "$0")/python" "$0" "$@"
' '''
import sys
from mesonbuild.mesonmain import main
if __name__ == '__main__':
    if sys.argv[0].endswith('.exe'):
        sys.argv[0] = sys.argv[0][:-4]
    sys.exit(main())
```

</details>

> this approach requires having /bin/sh available, may not be true e.g. in a Docker container-- using a binary-based wrapper could no longer be called with python directly

It means that prior this PR, this argument was already true, we needed `/bin/sh` for Meson, but not for AWS CLI. And now with this PR, we need it for both packages. I can see that the situation is not ideal. Would it make sense to you @jokeyrhyme to switch to `#!/usr/bin/env sh` as a first step ?

jokeyrhyme:
oh, I'm mostly just a brioche user, and haven't internalised a lot of the core concepts yet, so I'm mostly thinking out loud :) more for my own understanding than an actual suggestion 🤷

PR #1670, opened 2025-11-04
refactor(std): tweak live update scripts to be more idiomatic

jaudiger:
This PR tries to simplify a bit the Nushell scripts used for the live update, and try to be more idiomatic:

- Replace `($var | length) != 0` with `$var | is-not-empty`
- Replace `($var | length) == 0` with `$var | is-empty`
- Replace `get -o 0` with `into record`
- Streamline the filtering of versions with a `where` statement
- Use pipes when creating the record containing the version instead of using an intermediate variable 
- Reduce the scope of `mut project`


kylewlacy (approved):
LGTM :+1:

PR #1680, opened 2025-11-06
refactor(std): removes all the asert with Brioche version

jaudiger:
This PR tends to simplify, a bit, some of the std methods by removing the assert with old Brioche versions. They were asserts for:

- Brioche 0.1.0: 1
- Brioche 0.1.1: 2
- Brioche 0.1.2: 5
- Brioche 0.1.4: 1

[Brioche `0.1.4`](https://github.com/brioche-dev/brioche/releases/tag/v0.1.4) was released in January 2025, and [Brioche `0.1.2`](https://github.com/brioche-dev/brioche/releases/tag/v0.1.2) in September 2024.

I think (at least I hope) that most of the users now use the release of Brioche `0.1.5` or now `0.1.6`. I don't know if we have any statistics on that, I mean, a statistic that could show which version of Brioche was used when the registry was accessed (for example).

Maybe I'm too radical here, and I could re-add the assert to check until Brioche `0.1.4`, or maybe this PR is too early. What do you think @kylewlacy ?

### DIFF
--- a/packages/std/core/global.bri
+++ b/packages/std/core/global.bri
@@ -5,9 +5,6 @@ import {
   createRecipe,
 } from "./recipes";
 import { source } from "./source.bri";
-import { BRIOCHE_VERSION } from "./runtime.bri";
-import { semverMatches } from "./semver.bri";
-import { assert } from "./utils.bri";
 
 /**
  * Options for retrieving a git ref commit hash.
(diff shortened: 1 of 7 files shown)


kylewlacy (approved):
LGTM 🎉 

Yeah I think this refactoring makes sense. Old versions of Brioche can't even pull `std` from the registry today due to circular dependencies and the move to the new cache, so definitely in favor of getting rid of these extra checks!

PR #1682, opened 2025-11-06
New way to handle other versions

jaudiger:
It would have taken time, but here we go. This PR is here to open the discussion on how to handle multiple versions of a recipe. I'm not quite happy with what we were doing, and I'm still not happy with this proposal.

## Basic user story

Having a way to choose the version of a `recipe` when we want to install it locally or when we want to use it in an other recipes.

## How do we do it today ?

```ts
export const project = {
  name: "python",
  version: "3.14.0",
  repository: "https://github.com/python/cpython",
  extra: {
    otherVersions: {
      "3.14": "3.14.0",
      "3.13": "3.13.9",
      "3.12": "3.12.12",
    },
  },
};
```

Basically, we have `project.version` that serves as the latest version of a recipe, and we can have an optional key `project.extra.otherVersions` which exposes other possible versions that can be used for this recipe. Today, the concept of other versions could be defined as: "X(.Y)?" where X is the major number and Y the optional minor number. Here we don't want `Z` to be part of the tag.

## What we really want ?

This is not yet completely decided, but here is a [good summary](https://github.com/brioche-dev/brioche/discussions/358#discussioncomment-14849959) from @kylewlacy.

The CLI interface could be defined as is:

```bash
# From the registry
brioche install -r sqlx_cli@tag
```

Where `tag` is optional and could take one of these values:

 - `latest`
 - `X(.Y)?`
 - a hash

## Questions

- Do we really want to have a tag `latest` which is the same thing as just not setting the tag ?
- How do we handle it when installing from a local path ? We cannot use `@` since it's a valid character for a folder name

```bash
# From a local path
brioche install -p ./packages/sqlx_cli@tag
```

- Do we want to offer the same level of pinning when using a recipe in another recipe ? (aka `latest`, `X(.Y)?` and a hash)
- If we let choose a hash, how can we discover all the hashes for a recipe, and more importantly, how can we see the content of a recipe for a specific hash ?

### DIFF
--- a/packages/libxml2/project.bri
+++ b/packages/libxml2/project.bri
@@ -1,6 +1,6 @@
 import * as std from "std";
 import { cmakeBuild } from "cmake";
-import python, { getLatestVersion as getPythonVersion } from "python";
+import python, { getPythonVersion } from "python";
 import { nushellRunnable, type NushellRunnable } from "nushell";
 
 export const project = {
@@ -32,7 +32,7 @@ export default function libxml2(): std.Recipe<std.Directory> {
       // Remove it once the Python binding is suppressed (see https://gitlab.gnome.org/GNOME/libxml2/-/issues/891)
       LIBXML2_PYTHON_INSTALL_DIR: std.tpl`${
         std.outputPath
-      }/lib/python${getPythonVersion()}/site-packages`,
+      }/lib/python${getPythonVersion({ full: false })}/site-packages`,
       LIBXML2_WITH_TESTS: "OFF",
     },
   }).pipe((recipe) =>


jaudiger:
> For the final design, I still feel pretty happy overall with the design I outlined here: https://github.com/brioche-dev/brioche/issues/97#issuecomment-2232703901

The only downside of this proposal is that if a recipe depends on another recipe and we need to override its version as in the `aws_cli` recipe, it'll make things harder when the dependency is upsated:

Today, we have that:

```ts
export default function awsCli(): std.Recipe<std.Directory> {
  const pythonVersion = "3.13";
  const managedPython = python({ version: pythonVersion });

  let venv = std.recipe(managedPython);
}
```

With the proposal, we'll have:

```ts
export default function awsCli(): std.Recipe<std.Directory> {
  const pythonVersion = "3.13.9";
  const managedPython = python({ version: pythonVersion });

  let venv = std.recipe(managedPython);
}
```

Which means every time, the path version if Python changes, we need to update all the downstream recipes. I would more see something that is a mix of today and your proposal:

```ts
export const project = {
  name: "postgresql",
  version: "16.0.0",
  versions: [
    "15": "15.0.0",
    "14": "14.0.0",
  ],
};
```

But again with this proposal, it means the tag is `15`, `14` and `latest` (which redirects to `16.0.0`). But how to handle the tag `16` ? Should we put it in the key `versions` ? A bit as we have today (before this PR).

> Honestly, I would still push for using @. Yes, it can conflict with a directory name which is a tradeoff, but realistically most projects in the real world will use alphanumeric or -_ as names, and I think it's reasonable to enforce that if it's better ergonomics for the 99.9% of users who'd never even try to use @ or another special character in their folder name.

As far as [this character is forbidden in the name of a recipe](https://github.com/brioche-dev/brioche/issues/321), I guess, it can work. I agree that `@` is the most common character to represent what we want here.

> Shorter term, I think it'd be reasonable to add brioche subcommands to query the registry: search packages by name (or other fields), show info and versions for a package, fetch a specific version of a package, etc. This would require registry work too, since we don't have APIs for any of these yet though.

I agree for the MVP, we should have a way to request the available versions (and all its hashes ?) from the CLI. The webpage could come later.

> So I want to save the top-level otherVersions until we're super sure we know what we want the schema to be!

Sure, we can experiment in the `extra` key, and move the work in the root key once we're enough confident with the design.

> (same with latest too-- packages should ideally never need to see the word latest in their implementation)

Agree, that's why I wasn't feeling comfortable with the current approach of this PR.

> I think I would like to not merge this PR though (...unless we have a use-case where we need major.minor / latest tags)

Sure, I wasn't expected to be merged anytime soon ! I'll just extract the fix for the go recipe, plus a quick uniformization we could have.

jaudiger:
Also, after a second thought, it's maybe just me, but I'm not a big fan of having both `version` and `versions` keys in the project. Let's say we have nothing yet, and we can start from a blank page (just for the discussion). What would be the best way to support multiple versions while in the same time single version recipe ?

## 1

```ts
export const project = {
  name: "single",
  versions: [
    "1": "1.0.0",
  ],
};
```

```ts
export const project = {
  name: "multiple",
  versions: [
    "1": "1.0.0",
    "2": "2.0.0",
  ],
};
```

Cons:

- For `single` recipe, we shouldn't have a named tag (the only acceptable tag should be `latest`)
- For `multiple` recipe, the latest tag is a bit hidden. It'll be the highest number, but it's not obvious by just reading the recipe definition.

## 2

```ts
export const project = {
  name: "single",
  version: "1.1.0",
};
```

```ts
export const project = {
  name: "multiple",
  versions: [
    "1": "1.1.0",
    "2": "2.0.0",
  ],
};
```

Cons:

- For `multiple` recipe, the latest tag is a bit hidden. It'll be the highest number, but it's not obvious by just reading the recipe definition.

## 3

```ts
export const project = {
  name: "single",
  versions: "1.1.0",
};
```

```ts
export const project = {
  name: "multiple",
  versions: [
    "1.1.0",
    "2.0.0",
  ],
};
```

Cons:

- How to define named tags `1`, `2` for `multiple` recipe ?
- For `multiple` recipe, the latest tag is a bit hidden. It'll be the highest number, but it's not obvious by just reading the recipe definition.

## 4

```ts
export const project = {
  name: "single",
  version: "1.1.0",
};
```

```ts
export const project = {
  name: "multiple",
  versions: [
    "1.1.0",
    "2.0.0",
  ],
  tags: [
    "1": "1.1.0",
    "2": "2.0.0",
  ],
};
```

Cons:

- We have two keys in `multiple` recipe
- For `multiple` recipe, the latest tag is a bit hidden. It'll be the highest number, but it's not obvious by just reading the recipe definition.

## 5

Any other ideas ?

jaudiger:
I kind of like (4), since it's not a breaking change for 98% of current recipes.

For recipe with one version, we should only allow `latest` as tag, and `latest` will forward to `project.version`.

For recipe with multiple versions, `latest` will be the max of `project.versions`. And additional tags (numerical or alphabetical could be defined in `project.tags`, we could imagine something like that in the future:

```ts
export const project = {
  name: "multiple",
  versions: [
    "1.1.0",
    "2.0.0-rc2",
    "3.0.0-2025-11-01"
  ],
  tags: [
    "1": "1.1.0",
    "beta": "2.0.0-rc2",
    "nightly": "3.0.0-2025-11-01"
  ],
};
```

So it means a such recipe could be installed with these different commands:

```
brioche install multiple
brioche install multiple@latest
brioche install multiple@1
brioche install multiple@beta
brioche install multiple@nightly
```

kylewlacy:
@jaudiger I like the last example you gave, I think that's the one that resonates with me the most, but yeah it gets tricky when you start considering how to handle things like "setting `latest` by default" vs. "mixing stable + nightly versions". I like the idea of having `latest` default to the max version with the option to override it explicitly. Along the same lines, it might be good to automatically add the tags like `1.75` and `1` when the latest version is `1.75.2`, for example

...oh, and I can't remember if we ever discussed how the `import`s could look on the JS side before? e.g. for using the latest nightly version of Brioche, I think it'd be pretty nice to allow this:

```typescript
import rust from "rust@nightly";
```

...which means you wouldn't need to call it like `rust({ version: "nightly" })` or equivalent. Or, a similar option would be to use [import attribute syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with):

```typescript
import rust from "rust" with { version: "nightly" };
```

This gets somewhat more complicated to handle on the runtime side, but I feel like this syntax would make it pretty intuitive to use a specific pinned version of something!

PR #1719, opened 2025-11-07
ci: improve live update workflow when a branch already exists

jaudiger:
As discussed on Discord, this PR tries to improve live update workflow when a branch already exists by merging main into it.

I also merged the commands `git restore` and `git checkout` by using the flag `--force` from `git checkout`:

```
       -f, --force
           When switching branches, proceed even if the index or the working tree differs from HEAD, and even if there are untracked files in the way. This is used to throw away local changes and any untracked files or
           directories that are in the way.

           When checking out paths from the index, do not fail upon unmerged entries; instead, unmerged entries are ignored.
```

There is one difference from what I proposed on Discord, no comment is pushed if a PR is already opened, it was adding too much code for something that is maybe not worth it, right now. So, as for now, only a log is produced to inform a error was encountered.

### DIFF
--- a/.github/workflows/live-updates.yml
+++ b/.github/workflows/live-updates.yml
@@ -57,24 +57,24 @@ jobs:
             if grep -q 'export.*liveUpdate' "$package/project.bri"; then
               echo "::group::$label"
 
-              if git fetch --depth=1 origin "$branch_name"; then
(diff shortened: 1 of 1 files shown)


kylewlacy (approved):
LGTM

PR #1750, opened 2025-11-10
Add libimagequant package

jaudiger:
# Add a new Brioche recipe

Based on #1749 

## Summary

This Pull Request adds a new recipe to the registry.

- **Recipe name:** `libimagequant`
- **Website / repository:** https://pngquant.org/lib/
- **Short description:** Convert 24/32-bit images to 8-bit palette with alpha channel. C API/FFI libimagequant that powers pngquant lossy PNG compressor.

## Related issue(s) or discussion(s)

None.

## Checklist (required)

- [x] `liveUpdate()` method added.
- [x] `test()` method added.

## How I tested this locally (required)

**When applicable, the commands below must succeed and their output must be pasted into this PR.**

1. Run the **test** scenario:

```bash
brioche build -e test -p RECIPE_PATH
```

<details><summary>Test output (click to expand)</summary>
<p>

```
51556  │ 4.1.0
 0.05s ✓ Process 51556
Build finished, completed 1 job in 8.02s
Result: 03528a775a15e6a9a3d1f84a29052733505bcb268defadcc0aa7c4aacafc1c0f
```

</p>
</details>

2. Run the **live-update** scenario:

```bash
brioche run -e liveUpdate -p RECIPE_PATH
```

<details><summary>Live-update output (click to expand)</summary>
<p>

```
Build finished, completed 1 job in 12.12s
Running brioche-run
{
  "name": "libimagequant",
  "version": "4.1.0",
  "extra": {
    "crateName": "imagequant-sys"
  }
}
```

</p>
</details>

## Implementation notes / special instructions

- If this introduces breaking changes, list them and any required follow-ups.

### DIFF
--- a/packages/libimagequant/brioche.lock
+++ b/packages/libimagequant/brioche.lock
@@ -1,9 +1,8 @@
 {
   "dependencies": {},
-  "downloads": {
-    "https://crates.io/api/v1/crates/imagequant-sys/4.1.0/download": {
-      "type": "sha256",
-      "value": "b5361d7aab1d0c5623172cd6d18920bf21ef56f6e21d7290f1e31e0794c4a76e"
+  "git_refs": {
+    "https://github.com/ImageOptim/libimagequant.git": {
+      "4.4.1": "24e2956a37cd7ad1f4b81c0e20318e3239eb71dc"
     }
   }
 }
--- a/packages/libimagequant/project.bri
+++ b/packages/libimagequant/project.bri
@@ -3,22 +3,25 @@ import { cargoCBuild } from "cargo_c";
 
 export const project = {
   name: "libimagequant",
-  version: "4.1.0",
-  extra: {
-    crateName: "imagequant-sys",
-  },
+  version: "4.4.1",
+  repository: "https://github.com/ImageOptim/libimagequant.git",
 };
 
-const source = Brioche.download(
-  `https://crates.io/api/v1/crates/${project.extra.crateName}/${project.version}/download`,
-)
-  .unarchive("tar", "gzip")
-  .peel();
+const source = Brioche.gitCheckout({
+  repository: project.repository,
+  ref: project.version,
+});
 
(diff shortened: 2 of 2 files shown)


kylewlacy (approved):
LGTM

(Will wait until #1749 is merged to avoid potential conflicts)

kylewlacy:
Actually, dismissed my review after reading more on the version question

My gut instinct is that version `4.4.0` or `4.4.1` is the better choice here (not sure which yet between the two though...). To me, it seems like `libimagequant` is the project itself, where the C library (exposed as the `libimagequant-sys` crate) is a subproject-- effectively an artifact that's part of the overall project. Similar to how we don't have a `clang` package, it's just a part of the larger LLVM project.

The "rule of thumb" I've used is "are the projects part of a single repository"? If the answer is yes, I tend to feel that the whole repo should be the project, with different pieces exported... not a hard rule though: Google is a bit (in)famous for their monorepo setup and that's an example where it _probably_ shouldn't be just one big project! I think these cases are where I feel pretty strongly that we should be using `Brioche.gitCheckout` rather than downloading from `crates.io`.

kylewlacy:
Alright, looks like the upstream ticket was resolved, and there's now parity between latest git tag and `crates.io` release!  It sounds like treat the top-level project version as the "true" version... somewhat unfortunate with the difference from the `crates.io` release of the library, but this seems like a good reason to pull the source from GitHub in this case

kylewlacy:
Honestly, I'm fine with it either way! I'd lean towards getting this in just so we aren't blocked waiting for upstream, and we can add the improved tests as a follow-up. But if you'd prefer to wait so we can avoid an extra PR cycle, I'm fine with that too

jaudiger:
I'm fine to merge it as is, this package does not frequently release a new version. We can merge as is, and update the tests on next release or when we get feedbacks from upstream.

kylewlacy (approved):

PR #4157, opened 2026-04-17
fix: resolve eslint 9 issues

jaudiger:
Companion PR of https://github.com/brioche-dev/brioche/pull/452.

This PR resolves the following lints detected by ESLint 9:

```
[Warning] file:///tmp/brioche-packages/packages/std/toolchain/native/index.bri:347:12
Unnecessary conditional, comparison is always true, since `"path" === "path"` is true.

[Warning] file:///tmp/brioche-packages/packages/std/extra/autopack.bri:30:12
Unnecessary conditional, comparison is always true, since `"path" === "path"` is true.

[Warning] file:///tmp/brioche-packages/packages/std/core/recipes/process.bri:196:21
Unexpected iterable of non-Promise (non-"Thenable") values passed to promise aggregator.

[Warning] file:///tmp/brioche-packages/packages/std/core/recipes/process.bri:432:5
Unused eslint-disable directive (no problems were reported).

[Warning] file:///tmp/brioche-packages/packages/std/core/recipes/process.bri:439:9
Unexpected iterable of non-Promise (non-"Thenable") values passed to promise aggregator.

[Warning] file:///tmp/brioche-packages/packages/std/core/recipes/proxy.bri:46:5
Prefer using nullish coalescing operator (`??=`) instead of an assignment expression, as it is simpler to read.
```

### DIFF
--- a/packages/std/core/recipes/process.bri
+++ b/packages/std/core/recipes/process.bri
@@ -193,7 +193,8 @@ export function process(options: ProcessOptions): Process {
             await v.briocheSerialize(),
           ]),
         ),
-        Promise.all(dependencies.map(async (dep) => dep.briocheSerialize())),
+        // eslint-disable-next-line @typescript-eslint/await-thenable
+        Promise.all(dependencies.map((dep) => dep.briocheSerialize())),
       ]);
 
       return {
@@ -435,7 +436,8 @@ export class ProcessTemplate {
   briocheSerialize(): Promise<runtime.ProcessTemplate> {
     this.#serialized ??= (async () => {
       const components = await Promise.all(
-        this.components.map(async (component) =>
+        // eslint-disable-next-line @typescript-eslint/await-thenable
+        this.components.map((component) =>
           typeof component === "function" ? component() : component,
         ),
       );
--- a/packages/std/core/recipes/process.bri
+++ b/packages/std/core/recipes/process.bri
@@ -430,6 +430,9 @@ export class ProcessTemplate {
   #serialized: Promise<runtime.ProcessTemplate> | undefined;
 
   constructor(...components: ProcessTemplateLike[]) {
+    // NOTE: Needed to silence warnings for ESLint in Brioche v0.1.7
+    // and earlier.
+    // eslint-disable-next-line
     this.components = components;
   }
 
--- a/packages/std/core/recipes/process.bri
+++ b/packages/std/core/recipes/process.bri
@@ -193,7 +193,7 @@ export function process(options: ProcessOptions): Process {
             await v.briocheSerialize(),
           ]),
         ),
-        Promise.all(dependencies.map((dep) => dep.briocheSerialize())),
+        Promise.all(dependencies.map(async (dep) => dep.briocheSerialize())),
       ]);
 
       return {
@@ -429,14 +429,13 @@ export class ProcessTemplate {
   #serialized: Promise<runtime.ProcessTemplate> | undefined;
 
   constructor(...components: ProcessTemplateLike[]) {
-    /* eslint-disable-next-line */
     this.components = components;
   }
 
   briocheSerialize(): Promise<runtime.ProcessTemplate> {
     this.#serialized ??= (async () => {
       const components = await Promise.all(
-        this.components.map((component) =>
+        this.components.map(async (component) =>
           typeof component === "function" ? component() : component,
         ),
       );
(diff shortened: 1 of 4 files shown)


jaudiger:
It should be merged once the PR on Brioche is merged.

packages/std/core/recipes/process.bri
```diff
@@ -429,14 +429,13 @@ export class ProcessTemplate {
   #serialized: Promise<runtime.ProcessTemplate> | undefined;
 
   constructor(...components: ProcessTemplateLike[]) {
-    /* eslint-disable-next-line */
```
jaudiger, line 432: This comment is no more necessary with the upgrade.

packages/std/core/recipes/process.bri
```diff
@@ -193,7 +193,7 @@ export function process(options: ProcessOptions): Process {
             await v.briocheSerialize(),
           ]),
         ),
-        Promise.all(dependencies.map((dep) => dep.briocheSerialize())),
+        Promise.all(dependencies.map(async (dep) => dep.briocheSerialize())),
```
kylewlacy: I think for this one I would prefer to just use `eslint-disable-next-line`. I'm... not as familiar with how `async` works under-the-hood in JS runtimes, but my understanding is there is an extra cost to wrapping a function with `async` (but happy to be proven wrong on that!)

Since this is one of our most-used functions, I'd rather just silence the warning here.

```suggestion
        // eslint-disable-next-line @typescript-eslint/await-thenable
        Promise.all(dependencies.map(async (dep) => dep.briocheSerialize())),
```

packages/std/core/recipes/process.bri
```diff
@@ -429,14 +429,13 @@ export class ProcessTemplate {
...
   constructor(...components: ProcessTemplateLike[]) {
-    /* eslint-disable-next-line */
     this.components = components;
   }
 
   briocheSerialize(): Promise<runtime.ProcessTemplate> {
     this.#serialized ??= (async () => {
       const components = await Promise.all(
-        this.components.map((component) =>
+        this.components.map(async (component) =>
```
kylewlacy: Ditto for this one
```suggestion
        // eslint-disable-next-line @typescript-eslint/await-thenable
        this.components.map((component) =>
```

packages/std/extra/autopack.bri
```diff
@@ -25,12 +25,7 @@ export function autopack(
 
   const variableArgs: std.ProcessTemplateLike[] = Object.entries(
     variables,
-  ).flatMap(([name, value]) => {
-    switch (value.type) {
-      case "path":
-        return ["--var", std.tpl`${name}=path:${value.value}`];
-    }
-  });
+  ).flatMap(([name, value]) => ["--var", std.tpl`${name}=path:${value.value}`]);
```
kylewlacy: In this case, the use of a `switch` is meant to help us add other possible variants of `value.type` here. I really wish there were a way to not have this fire a lint warning, but anyway I'd prefer to just silence this one too

```suggestion
  ).flatMap(([name, value]) => {
    switch (value.type) {
      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
      case "path":
        return ["--var", std.tpl`${name}=path:${value.value}`];
    }
  });
```

packages/std/toolchain/native/index.bri
```diff
@@ -342,12 +342,7 @@ function autopack(
 
   const variableArgs: std.ProcessTemplateLike[] = Object.entries(
     variables,
-  ).flatMap(([name, value]) => {
-    switch (value.type) {
-      case "path":
-        return ["--var", std.tpl`${name}=path:${value.value}`];
-    }
-  });
+  ).flatMap(([name, value]) => ["--var", std.tpl`${name}=path:${value.value}`]);
```
kylewlacy: Ditto here

```suggestion
  ).flatMap(([name, value]) => {
    switch (value.type) {
      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
      case "path":
        return ["--var", std.tpl`${name}=path:${value.value}`];
    }
  });
```

kylewlacy (changes requested):
Ah thanks for getting this prepped in advance! Also happy to see this only affects the `std` package (it's a good sign that we keep weird stuff pretty well contained in `std` for the most part)

Honestly, I feel like all but one of the new lint warnings are more-or-less false positives (especially the `Promise.all` one feels like an oversight that it can't tell that a value _might_ be `Thenable`, which I think the function explicitly supports?). Anyway, I personally side towards silencing lint rules for those cases, since I felt the code quality or performance with the fixed versions was a small downgrade.

The `??=` one makes sense and is cleaner though!

jaudiger:
Got it, I reverted all the changes you requested. It should be ready for another review.

kylewlacy (approved):
LGTM

PR #4160, opened 2026-04-20
Enable GCC NLS support

jaudiger:
This PR enables Native Language Support (NLS) in the native GCC toolchain. Enabling NLS surfaced a latent bug: the native `gettext` was being built without `iconv()` support, breaking `msgfmt` for any package with non-UTF-8 `.po` files.

**During GCC toolchain build** (after flipping `--disable-nls` to `--enable-nls`):

```
msgfmt -o de.mo ../../../../libstdc++-v3/po/de.po
msgfmt: Cannot convert from "ISO-8859-1" to "UTF-8". msgfmt relies on iconv(). This version was built without iconv().
make[4]: *** [Makefile:564: de.mo] Error 1
```

**During `xkeyboard_config` build** (same root cause):

```
FAILED: [code=1] po/af/LC_MESSAGES/xkeyboard-config-2.mo
.../bin/msgfmt -o po/af/LC_MESSAGES/xkeyboard-config-2.mo ../po/af.po
.../bin/msgfmt: Cannot convert from "ISO-8859-1" to "UTF-8". msgfmt relies on iconv(). This version was built without iconv().
```

To be reviewed commit per commit:

- cccb304a: Root cause fix. Set `GCONV_PATH` during gettext's build so `configure` detects a working `iconv()` and `msgfmt` gets compiled with iconv support baked in.
- 86bb7486: Resolves the TODO: switch `--disable-nls` to `--enable-nls` and set `GCONV_PATH` so `msgfmt` can find gconv modules when compiling GCC's `.po` translation files.
- 943c614f: Expose `GCONV_PATH` via the toolchain's `setEnv` as a fallback, so downstream packages invoking `msgfmt` (meson's NLS handling, autotools, etc.) get charset conversion support automatically.
- cdbe7565: Drop the `nls: "false"` workaround now that the toolchain fix makes it unnecessary.

**Warning**: it will rebuild almost every recipe.

### DIFF
--- a/packages/ortp/brioche.lock
+++ b/packages/ortp/brioche.lock
@@ -1,7 +1,7 @@
 {
   "dependencies": {},
   "git_refs": {
-    "https://gitlab.linphone.org/BC/public/ortp.git": {
+    "https://github.com/BelledonneCommunications/ortp.git": {
       "5.4.97": "298490d88d7e531631a69fde96b4856272feb378"
     }
   }
--- a/packages/ortp/project.bri
+++ b/packages/ortp/project.bri
@@ -5,7 +5,9 @@ import bctoolbox from "bctoolbox";
 export const project = {
   name: "ortp",
   version: "5.4.97",
-  repository: "https://gitlab.linphone.org/BC/public/ortp.git",
+  // Use the GitHub mirror instead of the official GitLab repository,
+  // the later seems to have network issues.
+  repository: "https://github.com/BelledonneCommunications/ortp.git",
 };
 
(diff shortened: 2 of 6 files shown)


kylewlacy (approved):
Awesome, LGTM :+1:

Excited to support localization/translations/etc!

kylewlacy:
(Feel free to merge when ready, didn't want to queue it blindly in case there are still high-priority live updates, and because it's probably going to need some babysitting to re-queue it unfortunately)

PR #703, opened 2025-06-20
perf(git): use newer git clone argument to clone a single revision

jaudiger:
With the release of Git 2.49, we can rewrite our bash script to clone a repository at a specific revision without the history. See https://about.gitlab.com/blog/whats-new-in-git-2-49-0/#thin-clone-using---revision.

I benchmarked the new approach with `hyperfine`, and it's at least 1.1 times faster:

```bash
➜ export REPOSITORY="https://github.com/brioche-dev/brioche.git"
➜ export COMMIT="a4c817f0acc16536e4e528170bec0ee9782c4528"
➜ hyperfine \
  --prepare 'TMP=$(mktemp -d -p ./temp) && echo $TMP > .tmpdir' \
  --conclude 'rm -rf "$(cat .tmpdir)" && rm .tmpdir' \
  'git -c advice.detachedHead=false clone --depth 1 --revision=$COMMIT $REPOSITORY "$(cat .tmpdir)"' \
  'cd "$(cat .tmpdir)" && \
       git -c init.defaultBranch=main init && \
       git remote add origin $REPOSITORY && \
       git fetch --depth 1 origin $COMMIT && \
       git -c advice.detachedHead=false checkout FETCH_HEAD'
Benchmark 1: git -c advice.detachedHead=false clone --depth 1 --revision=$COMMIT $REPOSITORY "$(cat .tmpdir)"
  Time (mean ± σ):      1.236 s ±  0.085 s    [User: 0.106 s, System: 0.071 s]
  Range (min … max):    1.051 s …  1.346 s    10 runs

Benchmark 2: cd "$(cat .tmpdir)" && \
       git -c init.defaultBranch=main init && \
       git remote add origin $REPOSITORY && \
       git fetch --depth 1 origin $COMMIT && \
       git -c advice.detachedHead=false checkout FETCH_HEAD
  Time (mean ± σ):      1.391 s ±  0.093 s    [User: 0.025 s, System: 0.039 s]
  Range (min … max):    1.229 s …  1.536 s    10 runs

Summary
  git -c advice.detachedHead=false clone --depth 1 --revision=$COMMIT $REPOSITORY "$(cat .tmpdir)" ran
    1.13 ± 0.11 times faster than cd "$(cat .tmpdir)" && \
       git -c init.defaultBranch=main init && \
       git remote add origin $REPOSITORY && \
       git fetch --depth 1 origin $COMMIT && \
       git -c advice.detachedHead=false checkout FETCH_HEAD
```

PR #4648, opened 2026-06-11
Update `gitCheckout` functions to support SHA-256 git hashes

kylewlacy:
This PR updates the `gitCheckout` function from the `git` package to support checking out repos based on SHA-256 commit hashes. This function is also used by the `Brioche.gitCheckout` function from `std`, so this fix applies there as well.

While testing https://github.com/brioche-dev/brioche-packages/pull/703, I created [a small repo](https://codeberg.org/kylewlacy/git-repo-sha256)[^codeberg] and tried using it with this little example package:

```ts
import * as std from "std";

export default Brioche.gitCheckout({
  repository: "https://codeberg.org/kylewlacy/git-repo-sha256.git",
  ref: "main",
});
```

The lockfile updates properly, but it ends up failing because it trips [this assertion](https://github.com/brioche-dev/brioche-packages/blob/7c7e15a006c17bb365b2256f5a1e2c2efa070862/packages/git/project.bri#L127-L130), which checks for a 40-character hexadecimal hash (SHA-1). SHA-256 hashes are 64 hexadecimal characters.

Even after removing the assertion, it still fails because the `git clone --revision` command gives the following error:

```
Cloning into '/home/brioche-runner-c685f5bb060139f95244d738f99b0514c5c29a7412e43a5581bfad5ddab1c148/.local/share/brioche/outputs/output-c685f5bb060139f95244d738f99b0514c5c29a7412e43a5581bfad5ddab1c148'...
fatal: Remote revision a17d9e7867d105694f3c352f934ce8d326a3c7e61b4b3ab5d6c2a67e817406dd not found in upstream origin
```

I haven't looked into the cause of this issue at all, and I'm unsure if it's a limit of `git clone --revision` itself or if it's something on the remote's side (Codeberg).

Either way, I tried out the older method of cloning a single commit (prior to #703) and it seemed to work fine with SHA-256 hashes. So I decided to split based on whether we're cloning a SHA-1 hash or a SHA-256 hash, which I thought was pretty reasonable (even if it's a little annoying to have two different codepaths for cloning commits)!

[^codeberg]: Codeberg seems to be one of the few-- maybe the only?-- public forge that currently supports repos using the SHA-256 object format!

### DIFF
--- a/packages/git/project.bri
+++ b/packages/git/project.bri
@@ -123,9 +123,12 @@ export function gitCheckout(
     const { commit, repository, options: initOptions } = await init;
     options = { ...initOptions, ...options };
 
+    const isSha1Hash = /^[0-9a-f]{40}$/.test(commit);
+    const isSha256Hash = !isSha1Hash && /^[0-9a-f]{64}$/.test(commit);
+
     // Validate that the commit is a hash
     std.assert(
-      /^[0-9a-f]{40}$/.test(commit),
+      isSha1Hash || isSha256Hash,
       `Invalid git commit hash: ${commit}`,
     );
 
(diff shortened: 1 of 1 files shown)


jaudiger:
Indeed, GitHub does not support it yet (it's tracked here: https://github.com/orgs/community/discussions/154056). And I can confirm it locally:

```
✗ git remote add origin git@github.com:jaudiger/sha256-repo-test.git
git push -u origin main
fatal: the receiving end does not support this repository's hash algorithm
```

But we now have the building blocks to support it properly once the major forges will enable the feature.

jaudiger:
I think it'll take some more time for the tools and the forges to properly support SHA-256. I agree that having a second clone method here is not elegant. But let's merge it as is, and remove the hack once the situation is more stable upstream.

jaudiger (approved):

